PR #1, opened 2023-07-13
[SIW-277] Scaffold package for Wallet Instance Attestation

balanza:
#### List of Changes

* scaffold `io-react-native-wallet-instance-attestation` package

#### Motivation and Context

The `io-react-native-wallet-instance-attestation` will provide all data structures and workflows needed by the Wallet Instance to obtain and manage its attestation. 

The package is intended to work with React Native, hence it has been scaffolded using https://github.com/callstack/react-native-builder-bob. It comes with an example app that can be executed on both Android and iOS devices.

#### Checklist:


grausof (dismissed):
LGTM

balanza:
> LGTM. So I know I can be boring on this 🥹. What do you think about rename this package as `io-react-native-wia`?

I really have a thing against acronyms. If the problem is the length, I'd rather go for `io-wallet-instance-attestation` or `wallet-instance-attestation-rn`

hevelius:
> > LGTM. So I know I can be boring on this 🥹. What do you think about rename this package as `io-react-native-wia`?
> 
> I really have a thing against acronyms. If the problem is the length, I'd rather go for `io-wallet-instance-attestation` or `wallet-instance-attestation-rn`

What do you think about this? `io-rn-wallet-attestation`

balanza:
> > > LGTM. So I know I can be boring on this 🥹. What do you think about rename this package as `io-react-native-wia`?
> > 
> > 
> > I really have a thing against acronyms. If the problem is the length, I'd rather go for `io-wallet-instance-attestation` or `wallet-instance-attestation-rn`
> 
> What do you think about this? `io-rn-wallet-attestation`

Ehm... I don't want to be pedantic, but it's actually a wallet **instance** attestation.

Beside that... what are your concerns?

hevelius:
> > > > LGTM. So I know I can be boring on this 🥹. What do you think about rename this package as `io-react-native-wia`?
> > > 
> > > 
> > > I really have a thing against acronyms. If the problem is the length, I'd rather go for `io-wallet-instance-attestation` or `wallet-instance-attestation-rn`
> > 
> > 
> > What do you think about this? `io-rn-wallet-attestation`
> 
> Ehm... I don't want to be pedantic, but it's actually a wallet **instance** attestation.
> 
> Beside that... what are your concerns?

Ehm... of course I meant `io-rn-wallet-instance-attestation`. It could be probably a matter of style. Choosing a short package name can have several benefits for example in readability and consistency with other packages name. My preference however goes for `io-react-native-wia`. But again that's just my non-blocking opinion.

PR #10, opened 2023-07-25
feat: Add RP authentication

grausof:
Adds integration for authentication to RP.
The `Authorization Response Details` are missing

#### List of Changes
- Add `decodeAuthRequestQR`
- Add `getUnsignedWalletInstanceDPoP`
- Add `getRequestObject`

Unit tested

### DIFF
--- a/src/rp/index.ts
+++ b/src/rp/index.ts
@@ -13,12 +13,12 @@ import type { JWK } from "@pagopa/io-react-native-jwt/lib/typescript/types";
 export class RelyingPartySolution {
   relyingPartyBaseUrl: string;
   walletInstanceAttestation: string;
-  appFetch: GlobalFetch;
+  appFetch: GlobalFetch["fetch"];
 
   constructor(
     relyingPartyBaseUrl: string,
     walletInstanceAttestation: string,
-    appFetch: GlobalFetch = { fetch }
+    appFetch: GlobalFetch["fetch"] = fetch
   ) {
     this.relyingPartyBaseUrl = relyingPartyBaseUrl;
     this.walletInstanceAttestation = walletInstanceAttestation;
@@ -91,7 +91,7 @@ export class RelyingPartySolution {
     const decodedJwtDPop = await decodeJwt(signedWalletInstanceDPoP);
     const requestUri = decodedJwtDPop.payload.htu as string;
 
-    const response = await this.appFetch.fetch(requestUri, {
+    const response = await this.appFetch(requestUri, {
       method: "GET",
       headers: {
         Authorization: `DPoP ${this.walletInstanceAttestation}`,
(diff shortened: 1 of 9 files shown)


src/wallet-instance-attestation/issuing.ts
```diff
@@ -8,9 +8,13 @@ import { WalletInstanceAttestationIssuingError } from "../utils/errors";
 
 export class Issuing {
   walletProviderBaseUrl: string;
-
-  constructor(walletProviderBaseUrl: string) {
+  appFetch: GlobalFetch;
+  constructor(
+    walletProviderBaseUrl: string,
+    appFetch: GlobalFetch = { fetch }
```
balanza, line 14: **nitpick:** Plain args:
 ‏
```suggestion 
    appFetch: GlobalFetch["fetch"] =  fetch
```

balanza: (and of course it requires `this.appFetch.fetch` to become `this.fetch`)

src/utils/signature.ts
```diff
@@ -0,0 +1,4 @@
+import { SignJWT } from "@pagopa/io-react-native-jwt";
+
+export const getSignedJwt = async (unsignedJwt: string, signature: string) =>
+  await SignJWT.appendSignature(unsignedJwt, signature);
```
balanza, line 4: **question:** I'm honestly curious what's the value of this module, being it's just plain wrapper‏

src/rp/index.ts
```diff
@@ -0,0 +1,143 @@
...
+   * @function
+   * @param qrcode QR code content
+   *
+   * @returns The authentication request url
+   *
+   */
+  decodeAuthRequestQR(qrcode: string): string {
+    let decoded = decodeBase64(qrcode);
+    let decodedUrl = new URL(decoded);
+    let requestUri = decodedUrl.searchParams.get("request_uri");
```
balanza, line 39: Why `let`? (I also wonder why `eslint` doesn't complain, as we should have a rule to prefer `const`)

src/rp/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (requestUri) {
+      return requestUri;
+    } else {
+      throw new AuthRequestDecodeError(
```
balanza, line 43: **suggestion:** ‏Multiple steps of this method can fail. Should we wrap the whole body in a try/catch and raise a consistent error type?

balanza (approved):

src/wallet-instance-attestation/issuing.ts
grausof: Thanks, I'll fix it

src/utils/signature.ts
grausof: this wrapper avoids using `@pagopa/io-react-native-jwt` directly and installing it as a dependency. If you think it is superfluous we can eliminate it, np.

src/rp/index.ts
grausof: fixed

src/utils/signature.ts
balanza: Ah, I get it. I have no strong opinion about it. Would it be a problem for the app to **directly** reference the package? (it will be referenced anyway as dependency)

PR #12, opened 2023-07-25
feat: PID issuing [SIW-313]

grausof:
Adds the integration towards the PID issuer for obtaining a PID credential.

#### List of Changes
- Added `PAR` request with JWT generator
- Added `auth token` request with DPoP generator
- Added `credential` request with proof on nonce generator

#### Motivation and Context
This PR ports the SDK made only for Android into typescript code

Tested locally and through unit tests

#### How to use:
To understand how to get a PID refer to the `getPid` function within the sample app.

#### Checklist:

### DIFF
--- a/src/sd-jwt/index.ts
+++ b/src/sd-jwt/index.ts
@@ -36,6 +36,7 @@ export const decode = <S extends z.AnyZodObject>(
     header: decodedJwt.protectedHeader,
     payload: decodedJwt.payload,
   });
+  console.log(rawDisclosures);
 
   // get disclosures as list of triples
   // validate each triple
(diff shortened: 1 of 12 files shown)


src/sd-jwt/index.ts
```diff
@@ -36,6 +36,7 @@ export const decode = <S extends z.AnyZodObject>(
     header: decodedJwt.protectedHeader,
     payload: decodedJwt.payload,
   });
+  console.log(rawDisclosures);
```
hevelius, line 39: Is this console.log wanted?

hevelius (changes requested):

src/sd-jwt/index.ts
hevelius: the example app is on /example/src instead this is on /src

grausof: Oops 😳
Fixed!

hevelius (approved):
LGTM

PR #15, opened 2023-07-27
Fix QR code resolution

balanza:
#### List of Changes

* Define `QRCodePayload` type to map QR code content
* Polyfill `URL` class
* Add test scenario to example app

#### Motivation and Context

The main fix is to polyfill the `URL` class implementation using the `react-native-url-polyfill` package. This is needed because of a known, unsolved bug in React Native: https://github.com/facebook/react-native/issues/24428.

The other change is an improvement on the mapping of QR content, which is now a structured object containing all the information parsed from the received url.

Test scenario example app

#### Checklist:


grausof (approved):
LGTM

PR #17, opened 2023-07-27
chore: [SIW-271] Cross-device flow test in example app

balanza:
#### List of Changes

* add test case

#### Motivation and Context

Implement a test case for the [Cross-device flow](https://italia.github.io/eudi-wallet-it-docs/v0.4.1/en/relying-party-solution.html#remote-cross-device-flow)

The test case is incomplete as it ends with obtaining the request object from the Relying Party. Other steps to be implemented in next PRs

#### Checklist:

### DIFF
--- a/example/src/scenarios/cross-device-flow-with-rp.tsx
+++ b/example/src/scenarios/cross-device-flow-with-rp.tsx
@@ -46,7 +46,6 @@ export default async () => {
     // Scan/Decode QR
     const { requestURI: authRequestUrl } =
       RelyingPartySolution.decodeAuthRequestQR(QR);
-    //const authRequestUrl = "https://verifier.example.org/request_uri";
 
     // instantiate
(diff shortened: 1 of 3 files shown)


example/src/scenarios/cross-device-flow-with-rp.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export default async () => {
+  try {
+    // obtain new attestation
+    const WIA = await getAttestation();
+
+    // Scan/Decode QR
+    const { requestURI: authRequestUrl } =
+      RelyingPartySolution.decodeAuthRequestQR(QR);
+    //const authRequestUrl = "https://verifier.example.org/request_uri";
```
grausof, line 49: If you don't need it we can remove it!

grausof (approved):

PR #21, opened 2023-08-01
feat: [SIW-351] Encrypt authorization response

balanza:
#### List of Changes

* encrypt authorization response

#### Motivation and Context

As described in https://italia.github.io/eudi-wallet-it-docs/v0.4.1/en/relying-party-solution.html#authorization-response-details, authorization response payload must be encrypted in order to avoid SSL split attack vulnerabilities. 
Encryption has been made according to OpenID4VP specification https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-signed-and-encrypted-respon

#### Checklist:

### DIFF
--- a/src/rp/index.ts
+++ b/src/rp/index.ts
@@ -1,9 +1,14 @@
-import { AuthRequestDecodeError, IoWalletError } from "../utils/errors";
+import {
+  AuthRequestDecodeError,
+  IoWalletError,
+  NoSuitableKeysFoundInEntityConfiguration,
+} from "../utils/errors";
 import {
   decode as decodeJwt,
   decodeBase64,
   sha256ToBase64,
   SignJWT,
+  EncryptJwe,
 } from "@pagopa/io-react-native-jwt";
 import {
   QRCodePayload,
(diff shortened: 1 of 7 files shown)


src/rp/index.ts
```diff
@@ -206,6 +219,36 @@ export class RelyingPartySolution {
...
+  /**
+   * Select a public key from those provided by the RP
+   *
+   * @param entity The RP entity configuration
+   * @returns a
+   * @throws {NoSuitableKeysFoundInEntityConfiguration} If entity do not contain any public key suitable for encrypting
+   */
+  private choosePublicKeyToEncrypt(
+    entity: RpEntityConfiguration
+  ): JWK & { alg: "RSA-OAEP-256" | "RSA-OAEP" } {
```
grausof, line 231: what do you think about passing to this function the algorithm to use?

src/rp/index.ts
```diff
@@ -166,29 +171,37 @@ export class RelyingPartySolution {
...
       state: requestObj.payload.state,
       // TODO: [SIW-352] MUST add presentation_submission
       // presentation_submission:
       vp_token,
-    };
-    const formBody = new URLSearchParams({ response: authzResponse });
+    });
+    const encrypted = await new EncryptJwe(authzResponsePayload, {
+      alg: jwk.alg,
+      enc: "A256CBC-HS512",
```
grausof, line 201: alg is related to enc. With `A256CBC-HS512` you can use only `RSA-OAEP-256`. For this reason it is perhaps better to pass the algorithm to use to `choosePublicKeyToEncrypt`

grausof (changes requested):

src/rp/index.ts
balanza: Do we have to consider jwk on this choice? I'd rather use a fixed algorithm

grausof: You can use a fixed algorithm, the important thing is that JWK is compatible with the algorithm used.
With `RSA-OAEP-256` you can only use `A256CBC-HS512` and with  `RSA-OAEP` you can only use `A128CBC-HS256`.

in your code you fixed the `enc` at `A256CBC-HS512` but your JWK can be either of type `RSA-OAEP-256` and `RSA-OAEP`

grausof: my suggestion is to fix the key algorithm to `RSA-OAEP-256` with `A256CBC-HS512`. If then the key is not found then switch to `RSA-OAEP` with `A128CBC-HS256`. If that isn't found either, then an error returns

balanza: done in https://github.com/pagopa/io-react-native-wallet/pull/21/commits/8316b4fe3c23d8b236991ed8fa26f0ddab12ba5b

src/rp/index.ts
```diff
@@ -206,6 +219,51 @@ export class RelyingPartySolution {
...
+    | { jwk: JWK & { alg: "RSA-OAEP" }; enc: "A128CBC-HS256" } {
+    const shuffle = () => (Math.random() > 0.5 ? 1 : -1);
+
+    // Look for keys using "RSA-OAEP-256", and pick a random one
+    const [usingRsa256] = entity.payload.jwks.keys
+      .filter(
+        <T>(k: T & { alg?: string }): k is T & { alg: "RSA-OAEP-256" } =>
+          typeof k.alg === "string" && k.alg === "RSA-OAEP-256"
+      )
+      .sort(shuffle);
```
grausof, line 243: Why do you need it?
There is no need to pick a key randomly but by default just pick the first one that matches the requirement.

src/rp/index.ts
```diff
@@ -206,6 +219,51 @@ export class RelyingPartySolution {
...
+
+    if (usingRsa256) {
+      return { jwk: usingRsa256, enc: "A256CBC-HS512" };
```
grausof, line 246: since encryption has nothing to do with the key what do you think about moving this logic outside?

```suggestion
      return usingRsa256;
```

and use a function like:
```js
getEncryptionAlgByJwk(jwk: JWK): "A256CBC-HS512" | "A128CBC-HS256"
```

grausof (commented):
Good job. 
Just a few small suggestions

src/rp/index.ts
grausof: also the way the sort function is written I'm taking either the first or the last one and never anything in between

src/rp/index.ts
balanza: Fair enough, I will.

grausof (approved):

PR #22, opened 2023-08-03
feat: [SIW-352] Presentation submission

balanza:
#### List of Changes

* add `presentation_submission` on authorization response payload

#### Motivation and Context

Satisfy the specification given in https://italia.github.io/eudi-wallet-it-docs/v0.4.1/en/relying-party-solution.html

#### Checklist:


__mocks__/@pagopa/io-react-native-jwt/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  const g = x
+    .map(function (a) {
+      return String.fromCharCode(parseInt(a, 16));
+    })
+    .join("");
+
+  return btoa(g);
+}
+
+function removePadding(encoded: string): string {
```
grausof, line 19: What do you think about importing it from `io-react-native-jwt`?

src/rp/index.ts
```diff
@@ -165,7 +168,18 @@ export class RelyingPartySolution {
       })
       .toSign();
 
-    return vp_token;
+    const [, vc_scope] = requestObj.payload.scope;
+    const presentation_submission = {
+      definition_id: "32f54163-7166-48f1-93d8-ff217bdb0653",
```
grausof, line 173: According to https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#response-parameters I believe this may be a description of the type of presentation we are doing. For example `pid_presentation`

src/rp/index.ts
```diff
@@ -165,7 +168,18 @@ export class RelyingPartySolution {
...
+      id: "04a98be3-7fb0-4cf5-af9a-31579c8b0e7d",
```
grausof, line 174: This value should be randomized
```suggestion
      id: `${uuid.v4()}`,
```

src/sd-jwt/__test__/index.test.ts
```diff
@@ -91,37 +91,46 @@ describe("decode", () => {
...
 
     expect(result).toEqual(expected);
   });
 
   it("should fail on unknown claim", () => {
-    const fn = () => disclose(token, ["unknown"]);
+    const fn = disclose(token, ["unknown"]);
 
-    expect(fn).toThrow();
+    expect(fn).rejects.toBe({});
```
grausof, line 134: Check code review warning: `Async assertions must be awaited or returned`

grausof (changes requested):

src/sd-jwt/index.ts
```diff
@@ -61,27 +70,64 @@ export const decode = <S extends z.AnyZodObject>(
...
+  const paths = await Promise.all(
+    claims.map(async (claim) => {
+      const disclosure = disclosures.find(
+        ({ decoded: [, name] }) => name === claim
+      );
+      const hash = await sha256ToBase64(disclosure?.encoded || "");
+
+      // _sd is defined in verified_claims.claims and verified_claims.verification
+      // we must look into both
+      if (sdJwt.payload.verified_claims.claims._sd.includes(hash)) {
```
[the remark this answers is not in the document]
balanza: I think I didn't understand. This is not a fallback, we have to search in both sets - claims and verification. The point is, as claims are bigger and are more likely to be hit, we perform the search sequentially (is it in a claim? If no, it must be in verification...)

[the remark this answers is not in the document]
grausof: sorry, maybe I didn't explain myself well. I was referring to the empty string fall back option on line 111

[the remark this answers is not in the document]
balanza: It's just impossible to happen (but I didn't bother to instruct the type system accordingly). 

It's very ugly but it has an advantage: it does not pollute the flow with an option impossible to happen. But I'll think on a more elegant solution.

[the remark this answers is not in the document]
grausof: yes, if it's impossible to happen then better find a more elegant solution. Otherwise NP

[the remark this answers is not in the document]
balanza: Solved in https://github.com/pagopa/io-react-native-wallet/pull/22/commits/682dc52f0542632ce7f04e32a84a44ce2ec4c262

I decide to do the claim presence check in the same iteration of the path sought. So that we are sure the disclosure exists and we save one loop.

src/rp/index.ts
balanza: As per the provided example in https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html, I think we can use the first part of the scope field in the request object. Such a field I think is in the format `definition_id vc1 vc2 ... vcN` (not clear in specs).

Done in https://github.com/pagopa/io-react-native-wallet/pull/22/commits/c2ef778ec4a7ac83f893cf2b50a77ec62911db05

__mocks__/@pagopa/io-react-native-jwt/index.ts
balanza: I don't think it's an exported method

grausof: Ok, no problem

src/rp/index.ts
balanza: Done in https://github.com/pagopa/io-react-native-wallet/pull/22/commits/4a496e41c67328e29f1724bee30e7451c833b242

grausof: uuid.v4() is of type number[]. I think you have to use a string. Also you imported `uuid` twice

src/sd-jwt/__test__/index.test.ts
balanza: Tests have been fixed in https://github.com/pagopa/io-react-native-wallet/pull/24, I think we can merge that before

Depends on https://github.com/pagopa/io-react-native-wallet/pull/24

src/rp/index.ts
balanza: Done in https://github.com/pagopa/io-react-native-wallet/pull/22/commits/b31fb552ee275a7149983fc73c1ecbd8d316eaa1

grausof (approved):
Great!

PR #25, opened 2023-08-09
chore: Add function to obtain PID issuer metadata [SIW-396]

grausof:
Adds function below Issuer for getting metadata.

#### List of Changes
- Added `PID.Issuer.getMetadata()`

Unit tested

#### Checklist:

### DIFF
--- a/src/pid/issuing.ts
+++ b/src/pid/issuing.ts
@@ -10,7 +10,7 @@ import uuid from "react-native-uuid";
 import { PidIssuingError, PidMetadataError } from "../utils/errors";
 import { getUnsignedDPop } from "../utils/dpop";
 import { sign, generate, deleteKey } from "@pagopa/io-react-native-crypto";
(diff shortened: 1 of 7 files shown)


src/pid/issuing.ts
```diff
@@ -302,4 +304,39 @@ export class Issuing {
...
   }
+
+  /**
+   * Obtain the PID issuer metadata
+   *
+   * @function
+   * @returns PID issuer metadata
+   *
+   */
+  async getMetadata(): Promise<PidIssuerMetadata> {
```
balanza, line 315: **suggestion(naming):** Shouldn't it be `getEntityConfiguration`?‏

balanza (commented):
LGTM 🚀

src/pid/issuing.ts
grausof: yep! fixed!

balanza (approved):

PR #31, opened 2023-08-31
fix: [SIW-423] Fixes for integrating with Relying Party

balanza:
#### List of Changes

_**things we needed to change:**_
* use base64url encoding instead of base64 when needed
* passed correct values for DPoP public key, so that it matches the public key in the Wallet Instance Assertion
* passed `nonce` coming from Request Object in the credential presentation, along with unique `jti` and `definition_id`
* update test scenario where needed

**_temporary workarounds:_**
* loosen schema for RP's Entity Configuration, so we can deal with missing `federation_entity` field
* loosen schema for RP's Request Object, so we can deal with missing `typ` field
* loosen schema for RP's QRCode payload, so we can deal handle urls with any protocol (`eudiw` expected)

#### Motivation and Context

Some needed changes emerged while testing the integration with the test Relying Party implementation.

Example app

#### Checklist:


src/rp/index.ts
```diff
@@ -225,94 +233,72 @@ export class RelyingPartySolution {
...
   /**
    * Obtain the relying party entity configuration.
    */
   async getEntityConfiguration(): Promise<RpEntityConfiguration> {
-    const wellKnownUrl = new URL(
-      "/.well-known/openid-federation",
-      this.relyingPartyBaseUrl
-    ).href;
+    const wellKnownUrl =
+      this.relyingPartyBaseUrl + "/.well-known/openid-federation";
```
grausof, line 303:
```suggestion
      new URL("/.well-known/openid-federation", this.relyingPartyBaseUrl).href;
```

src/utils/jwk.ts
```diff
@@ -37,3 +37,26 @@ export const JWK = z.object({
...
+ * @see https://datatracker.ietf.org/doc/html/rfc7517
+ *
+ * @param key The key to fix
+ * @returns THe same input key with fixed values
+ */
+export function fixBase64EncodingOnKey(key: JWK): JWK {
+  const { x, y, e, n, ...pk } = key;
+  const removePadding = (encoded: string) =>
+    // eslint-disable-next-line no-div-regex
+    encoded.replace(/=/g, "").replace(/\+/g, "-").replace(/\//g, "_");
```
grausof, line 53: You can use `removePadding` from `io-react-native-jwt`

src/utils/jwk.ts
```diff
@@ -37,3 +37,26 @@ export const JWK = z.object({
...
+
+/**
+ * Ensure key values are encoded using base64url and not just base64, as defined in https://datatracker.ietf.org/doc/html/rfc7517
+ *
+ * @see https://datatracker.ietf.org/doc/html/rfc7517
+ *
+ * @param key The key to fix
+ * @returns THe same input key with fixed values
+ */
+export function fixBase64EncodingOnKey(key: JWK): JWK {
```
grausof, line 49: what do you think about adding this function to `io-react-native-jwt` ?

src/wallet-instance-attestation/issuing.ts
```diff
@@ -38,7 +38,7 @@ export class Issuing {
       jti: `${uuid.v4()}`,
       type: "WalletInstanceAttestationRequest",
       cnf: {
-        jwk: publicKey,
+        jwk: fixBase64EncodingOnKey(publicKey),
```
grausof, line 41: IMHO this parsing has to be done upstream by who generates the key in JWK format, so probably `io-react-native-crypto`

src/rp/index.ts
```diff
@@ -180,18 +183,25 @@ export class RelyingPartySolution {
...
+      iss: walletInstanceIdentifier,
+      nonce: requestObj.payload.nonce,
+    })
       .setAudience(requestObj.payload.response_uri)
+      .setIssuedAt()
       .setExpirationTime("1h")
       .setProtectedHeader({
         typ: "JWT",
         alg: "ES256",
+        kid: "",
```
grausof, line 198: Here the KID of the key with which the vp token will be signed must be entered. It can not be empty!

src/rp/index.ts
```diff
@@ -180,18 +183,25 @@ export class RelyingPartySolution {
...
+      jti: `${uuid.v4()}`,
+      iss: walletInstanceIdentifier,
+      nonce: requestObj.payload.nonce,
+    })
       .setAudience(requestObj.payload.response_uri)
+      .setIssuedAt()
       .setExpirationTime("1h")
       .setProtectedHeader({
         typ: "JWT",
         alg: "ES256",
```
grausof, line 197: I don't want to be picky, given that the same problem also exists for obtaining the PID. But this value cannot be static but depends on the key that will be used for signing. We need an inverse function of [getKtyFromAlg](https://github.com/pagopa/io-react-native-jwt/blob/0cc8243be6f6093ebf11f7f36cc6375ce3cc997e/src/algorithms.ts#L18)

grausof (changes requested):

src/utils/jwk.ts
balanza: I was thinking more on `io-react-native-crypto`. This is actually a deviation from the specification of any key in jwk format, it may concern all keys in the app. 

I'd leave it here for now

src/rp/index.ts
balanza: Is this true for every signed jwt?

src/wallet-instance-attestation/issuing.ts
grausof: ok

src/rp/index.ts
grausof: yep!

grausof:
Please check code review!

src/rp/index.ts
balanza: This is an issue occurring on multiple parts of the codebase like https://github.com/pagopa/io-react-native-wallet/commit/3683bcab054455459f239a6c59ec49a25532310e or https://github.com/pagopa/io-react-native-wallet/commit/0717493907935f6f9230b029642d160f15f6298d.

I made up my mind we need to handle the signing process in the `SignJWT` class, so that we don't have to pass intermediate values around the code. But it's just an idea.

I think we must handle the issue, but it's better we make a dedicated PR

src/rp/index.ts
balanza: Done in https://github.com/pagopa/io-react-native-wallet/pull/31/commits/dced748fd8a931965ce63ae62aa76fc76d422afa

balanza: The base url must be taken as-is, at it may have _path_ part. `URL` class will use only the hostname

Example:
```ts
new URL("/foo/bar", "https://example.com/base/path").href
// https://example.com/foo/bar
```

src/utils/jwk.ts
grausof: You can use `io-react-native-jwt` [0.6.4](https://github.com/pagopa/io-react-native-jwt/releases/tag/v0.6.4) which now exposes removePadding and addPadding

balanza: done in https://github.com/pagopa/io-react-native-wallet/pull/31/commits/f1f2dc2325dd5784529d52bd5ddc0422892f5758

src/wallet-instance-attestation/types.ts
```diff
@@ -60,7 +64,7 @@ export const WalletInstanceAttestationJwt = z.object({
       tos_uri: z.string().url(),
       logo_uri: z.string().url(),
       asc: z.string(),
-      authorization_endpoint: z.string().url(),
+      authorization_endpoint: z.string() /* .url() */,
```
grausof, line 67: is this comment helpful?

grausof (approved):
LGTM

src/wallet-instance-attestation/types.ts
balanza: removed

PR #34, opened 2023-09-08
feat: [SIW-402] Resolve trust chain

balanza:
#### List of Changes

* introduced `trust` folder for utilities related to the trust model
* Added/refined types for federaration entities
* Implemented trust chain verification
* added utility to fetch and parse Entity Configuration of federation entities
* added trust chain verification on example app

#### Motivation and Context

Documents shared with other parties (credentials, attestations, authz requests, etc) are signed and trusted according to https://italia.github.io/eudi-wallet-it-docs/v0.4.1/en/trust.html. Such documents contain a `trust_chain` field, which is intended to provide a static representation of the federation trust chain that connects the Leaf entity (i.e. the entity whose trustworthiness is being checked) to the known Trust Anchor entity,

The `trust_chain` is a list of signed token in the form:
```
[
  leaf entity configuration token,
  leaf entity statement token,
  intermediate1 entity statement token,
  ...
  intermediateN entity statement token,
  (optional) trust anchor entity configuration token,
]
```
Each token MUST be signed using one of the keys declared in the token next to it in the chain. The last token's signature will be checked against the known Trust Anchor's keys. Optionally, Trust Anchor entity configuration can be added as last element of the chain.

This PR proposes the `verifyTrustChain` function to be used as
```ts
import { getEntityConfiguration, verifyTrustChain } from "@pagopa/io-react-native-wallet";

const trustAnchorEC = await getEntityConfiguration("https://trustanchor.example");
const document = // the document received from the issuer/relying party, whose trust must be verified

const chain = await verifyTrustChain(trustAnchor, document.header.trust_chain);
```

#### Checklist:


src/rp/types.ts
```diff
@@ -28,41 +29,21 @@ export const RequestObject = z.object({
...
       }),
-      // FIXME: SIW-422 require federation_metadata field
-      // Actual RP implementation does not comply with the spec
-      /* federation_entity: z.object({
-        organization_name: z.string(),
-        homepage_uri: z.string(),
-        policy_uri: z.string(),
-        logo_uri: z.string(),
-        contacts: z.array(z.string()),
-      }), */
```
grausof, line 61: Sorry for rework but `federation_entity` has been added!

example/src/scenarios/cross-device-flow-with-rp.tsx
```diff
@@ -80,7 +90,7 @@ export default async () => {
     ).then((t) => RP.getRequestObject(t, authRequestUrl, entity));
 
     // Attest Relying Party trust
-    // TODO [SIW-354]
+    await verifyTrustChain(trustAnchorEntity, requestObj.header.trust_chain);
```
grausof, line 93: what do you think about adding this check also into the SDK function (like `getEntityConfiguration()`)making sure that the caller is not responsible for doing the checks.

src/trust/chain.ts
```diff
@@ -0,0 +1,109 @@
...
+// The first element of the chain is supposed to be the Entity Configuration for the document issuer
+const FirstElementShape = EntityConfiguration;
+// Each element but the first is supposed to be an Entity Statement
+const MiddleElementShape = EntityStatement;
+// The last element of the chain can either be an Entity Statement
+//  or the Entity Configuration for the known Trust Anchor
+const LastElementShape = z.union([
+  EntityStatement,
+  TrustAnchorEntityConfiguration,
+]);
```
grausof, line 49: What do you think about making the control programmatic? A priori we do not know how many entities there are in the chain. It could be 2,3 or 4.

src/trust/chain.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  // Select the expected token shape
+  const selectTokenShape = (elementIndex: number) =>
+    elementIndex === 0
+      ? FirstElementShape
+      : elementIndex === chain.length - 1
+      ? LastElementShape
+      : MiddleElementShape;
+
```
grausof, line 75: ditto

src/trust/types.ts
```diff
@@ -0,0 +1,58 @@
...
+    metadata: z.object({
+      // FIXME: SIW-422 require federation_metadata field
+      // Actual RP implementation does not comply with the spec
+      /* federation_entity: z.object({
+        organization_name: z.string(),
+        homepage_uri: z.string(),
+        policy_uri: z.string(),
+        logo_uri: z.string(),
+        contacts: z.array(z.string()),
+      }), */
```
grausof, line 49: RP added `federation_entity`. You can uncomment this part

grausof (changes requested):

src/trust/chain.ts
balanza: I don't understand the question. Those are utility types meant for a better undestand of chain structure. It allows for chains of any shape like

```ts
[FirstElementShape, LastElementShape]
[FirstElementShape, MiddleElementShape, LastElementShape]
[FirstElementShape, MiddleElementShape, MiddleElementShape, LastElementShape]
[FirstElementShape, MiddleElementShape, MiddleElementShape, MiddleElementShape, LastElementShape]
[FirstElementShape, MiddleElementShape, ... , MiddleElementShape, LastElementShape]
... and so on
```

The control is then programmatic, as it iterates through the actual chain to be verified, expecting each element to be in shape according to its position in the chain. (see `selectTokenShape`)

example/src/scenarios/cross-device-flow-with-rp.tsx
balanza: Interesting point. That is true for every trustable document we have (including PID, WIA, RequestObject, etc).

I think it's good, anyway I'd do it in another PR as it involves a sensible refactor of classes because:
* we must provide Trust Anchor data 
* we must model better error to avoid to hide informations on error.

grausof: yep! Thanks you

grausof (approved):
Please fix test since code review fail!

src/rp/types.ts
balanza: done in https://github.com/pagopa/io-react-native-wallet/pull/34/commits/98ca656f5398a5358b88769fd7c69afab8cf2203

grausof: Sorry but I still don't see the update [here](https://github.com/pagopa/io-react-native-wallet/blob/SIW-402--resolve-trust-chain/src/rp/types.ts)

grausof:
It is necessary to add (certainly with a separate PR) fast renewal for entity configurations.

grausof (approved):
LGTM

PR #35, opened 2023-09-08
chore: Fix linter

balanza:
#### List of Changes

* add `./example/src` as eslint target

#### Motivation and Context

`yarn lint` command will address both `./src` and `./example/src`

#### Checklist:


hevelius (approved):
LGTM

PR #38, opened 2023-09-14
Fix #34 Entity Configuration parsing

balanza:
#### List of Changes

* fix `EntityConfiguration` and `RpEntityConfiguration` parsers so they don't strip unknown keys
* set `federation_entity` attributes as optional, according to https://openid.net/specs/openid-connect-federation-1_0.html#appendix-A.2.2-5
* skip trust chain verification as it's not resolved yet by the test Relying Party
* expose `EntityConfiguration`, `TrustAnchorEntityConfiguration` and `RpEntityConfiguration` types

#### Motivation and Context

Bugs were introduced in https://github.com/pagopa/io-react-native-wallet/pull/34 as the entity configurations failed to be parsed. This is due to a misinterpreted feature of `zod` that [automatically strips additional fields on objects](https://github.com/colinhacks/zod/tree/v2#migration-from-v1).

Example app

#### Checklist:

### DIFF
--- a/example/src/scenarios/cross-device-flow-with-rp.tsx
+++ b/example/src/scenarios/cross-device-flow-with-rp.tsx
@@ -1,14 +1,14 @@
 import { sign, generate, getPublicKey } from "@pagopa/io-react-native-crypto";
-import { RelyingPartySolution } from "@pagopa/io-react-native-wallet";
 import {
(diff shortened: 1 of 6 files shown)


grausof (approved):
LGTM

PR #40, opened 2023-09-18
chore! [SIW-439] Update JWT signing

balanza:
#### List of Changes

See commit list

#### Motivation and Context

Receives the changes in the latest version of `@pagopa/io-react-native-jwt` which changes how cryptographic assets are handled when signing tokens (see https://github.com/pagopa/io-react-native-jwt/pull/21). 
As there is no more the need to sign tokens _outside_ the package, the API results simpler and with fewer operations for the app to do.
API also evolves in #41 

Example app

#### Checklist:

Depends on #41


example/src/scenarios/get-pid.tsx
```diff
@@ -1,91 +1,60 @@
...
 
-    //Generate nonce proof
-    const unsignedNonceProof = await issuingPID.getUnsignedNonceProof(
-      authToken.c_nonce
+    // throw if decode fails
+    const pidJwt = PID.SdJwt.decode(pid.credential);
+    console.log(
+      "PID holder binding JWK: ",
+      await thumbprint(pidJwt.sdJwt.payload.cnf.jwk)
     );
```
grausof, line 52:
```suggestion

```

example/src/scenarios/get-pid.tsx
```diff
@@ -1,91 +1,60 @@
...
-      authToken.access_token,
-      {
-        birthDate: "01/01/1990",
-        fiscalCode: "AAABBB00A00A000A",
-        name: "NAME",
-        surname: "SURNAME",
-      }
+    console.log(
+      "PID public key: ",
+      await thumbprint(await pidCryptoContext.getPublicKey())
```
grausof, line 55: I think we can delete it!

package.json
grausof, line 1: You need to increase the package version

example/src/scenarios/get-pid.tsx
```diff
@@ -1,90 +1,55 @@
...
+    const pid = await credentialRequest(authConf, pidEntityConfiguration, {
+      birthDate: "01/01/1990",
+      fiscalCode: "AAABBB00A00A000A",
+      name: "NAME",
+      surname: "SURNAME",
+    });
 
     // throw if decode fails
-    PID.SdJwt.decode(pid.credential);
+    const pidJwt = PID.SdJwt.decode(pid.credential);
```
grausof, line 48:
```suggestion

```

package.json
balanza: Done in https://github.com/pagopa/io-react-native-wallet/pull/40/commits/81b989dd5df9991d744fcc5f52234ce97536cb83. Let's stick with `0.x.x` major until it's unstable.

grausof (approved):

PR #41, opened 2023-09-18
chore!: [SIW-439, 496] Classes refactor

balanza:
Integrates changes proposed in https://github.com/pagopa/io-react-native-wallet/pull/40 with an API refactory.

#### List of Changes

* use functions instead of classes

#### Motivation and Context

When coding an user flow (e.g. PID issuing), our previous approach was to instantiate an instance for each flow and then call the instance's methods for each step:
```ts
// pseudo 
const flow = new MyFlow(/* parameters for the entire flow */);
flow.step1();
flow.step2();
```
With this PR we refactor the interface to use a _pure_ function for each step:
```ts
// pseudo 
MyFlow.step1(/* parameters for step1 only */));
MyFlow.step2(/* parameters for step2 only */));
```
The main purpose of the refactor is explicity:
* parameters are declared for each step, so their scope is well-defined
* there is no hidden, internal state of the instance that might be misunderstood by users
* data is passed as input/output from one step to another, making them callable in different places in application's code

#### Checklist:


grausof (approved):
Great 🎉

PR #36, opened 2023-09-08
refactor: [SIW-477] Migrate example app from Expo to bare RN

LazyAfternoons:
#### List of Changes
- Migrates the example app from Expo to bare React Native;
- Removes any Expo related dependency;
- Updates the RN version to the `0.72.4` version;
- Replaces the `android` and `ios` project folders with bare RN projects, scaffolded with `react-native-builder-bob`;
- Removes `webpack` and thus web support;
- Updates Babel, Metro and React Native configs as explained [here](https://github.com/callstack/react-native-builder-bob#how-does-the-library-get-linked-to-the-example-app-in-the-generated-project) in order to link the library to the example app and to the `node_modules` folder contained in the library directory.; 
- Adds `.node-version` and `.ruby-version` to align and manage environments of the example app;
- Updates the `README.md` of the example with better instructions on how to build the project.

#### Motivation and Context

The updated example app now aligns more closely with the current io-app environment. By removing Expo, we can effectively identify and troubleshoot bugs that may impact React Native in a bare project but not in an Expo environment. This change also offers the advantage of being able to build both apps on physical devices without relying on Expo CLI or an Expo account.

#### How Has This Been Tested?

Clone this repository from scratch and follow the instructions of the README. 
Everything should work as before.
Also test if any change to the library is immediately reflected while running the example app.

#### Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #42, opened 2023-09-19
build: [SIW-494] Remove old example xcworkspace

LazyAfternoons:
#### List of Changes
This pull request eliminates the outdated example xcworkspace which is no longer needed.

#### Motivation and Context
When attempting to execute `yarn ios` on the example app, React Native throws an error due to the obsolete xcworkspace, resulting in a failure to locate a project with that particular name. This problem arises as a consequence of the refactoring carried out in pull request #36. However, by removing the outdated xcworkspace, React Native will correctly interpret the xcworkspace with the updated name, `IoReactNativeWalletExample.xcworkspace`.

Run `yarn ios` and the example app should now build without throwing an error.

#### Checklist:


hevelius (approved):
LGTM. It works.

PR #45, opened 2023-09-29
feat: [SIW-559] Add deep link handler for cross device scenario in example App

grausof:
Adds integration to the `eudiw://` deeplink to test the same device authentication scenario towards RP.
⚠️ Warning: Only for the example app a deep link (eudiw://) was used but for future versions a universal link will be used as decided within the RFC.

#### List of Changes
- Added handler for `eudiw://` deep link which automatically starts the authentication flow towards RP.

#### Motivation and Context
This PR was necessary as it is no longer possible to use dated **QR Codes** (therefore hard coded in the app) as they are expired but a fresh session must be used. The implementation of the **same device flow** allows you to avoid using the camera on the simulator to obtain the content of the QR Code.

Tested on real devices and simulators on iOS and Android

https://github.com/pagopa/io-react-native-wallet/assets/442709/8cd1c17c-fe4a-4985-bdb9-4311c32db1eb

#### Checklist:

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,8 +1,39 @@
-import * as React from "react";
-import { StyleSheet, View, Text, Button, SafeAreaView } from "react-native";
+import {
+  StyleSheet,
+  View,
+  Text,
+  Button,
+  SafeAreaView,
+  Linking,
+} from "react-native";
 import scenarios, { type ScenarioRunner } from "./scenarios";
+import React, { useEffect } from "react";
+import "react-native-url-polyfill/auto";
+import { encodeBase64 } from "@pagopa/io-react-native-jwt";
 
 export default function App() {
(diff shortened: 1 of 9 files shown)


hevelius (approved):
LGTM. It works as expected.

example/src/App.tsx
```diff
@@ -1,19 +1,47 @@
...
+  useEffect(() => {
+    Linking.getInitialURL()
+      .then((url) => {
+        if (url !== null) {
+          setDeeplink(url);
+        }
+      })
+      .catch((err) => console.error("An error occurred", err));
+    let subcribtion = Linking.addEventListener("url", handleOpenURL);
+    subcribtion.subscriber;
```
balanza, line 26: ?

grausof: typo, fixed!

PR #47, opened 2023-10-03
feat!: [SIW-545] Adds the authorization endpoint for PID issuing

grausof:
Adds the call to the authorization endpoint to obtain the PID.

#### List of Changes
- Introduces a breaking change to `PID.Issuing.getCredential` and `PID.Issuing.authorizeIssuing`. In particular, the user data must be passed to `authorizeIssuing` and no longer to `getCredential`.
- Adds support for form_post as defined in https://github.com/italia/eudi-wallet-it-docs/issues/132

#### Motivation and Context
See https://github.com/italia/eudi-wallet-it-docs/issues/132

Unit tested and with simulator

#### Checklist:

### DIFF
--- a/src/pid/issuing.ts
+++ b/src/pid/issuing.ts
@@ -164,6 +164,11 @@ const getAuthenticationRequest =
       pidProviderEntityConfiguration.payload.metadata.openid_credential_issuer
         .authorization_endpoint;
 
+    /* User's personal data is not supposed to transit in this flow,
+     * but to be provided to the PID issuer directly by its chosen authentication method (CIE).
+     * Being the project in an initial phase, and being we were still unable to fully comply with authentication,
+     * we temporarily provide data from the App's logged user.
+     * */
     const params = new URLSearchParams({
       client_id: clientId,
       request_uri: requestUri,
--- a/src/pid/issuing.ts
+++ b/src/pid/issuing.ts
@@ -195,8 +195,9 @@ const getAuthenticationRequest =
  * @param params.wiaCryptoContext The key pair associated with the WIA. Will be use to prove the ownership of the attestation.
  * @param params.appFetch (optional) Http client
  * @param walletInstanceAttestation Wallet Instance Attestation token.
- * @param walletProviderBaseUrl Base url for the Wallet Provider
+ * @param walletProviderBaseUrl Base url for the Wallet Provider.
  * @param pidProviderEntityConfiguration The Entity Configuration of the PID Provider, from which discover public endooints.
+ * @param cieData Data red from the CIE login process
  * @returns The access token along with the values that identify the issuing session.
  */
 export const authorizeIssuing =
@@ -321,7 +322,6 @@ const createNonceProof = async (
  * @param params.pidCryptoContext The key pair associated with the PID. Will be use to prove the ownership of the credential.
  * @param params.appFetch (optional) Http client
  * @param authConf The authorization configuration retrieved with the access token
- * @param cieData Data red from the CIE login process
  * @returns The PID credential token
  */
 export const getCredential =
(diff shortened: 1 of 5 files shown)


src/pid/issuing.ts
```diff
@@ -135,14 +149,55 @@ const getPar =
...
+        .authorization_endpoint;
+
+    const params = new URLSearchParams({
+      client_id: clientId,
+      request_uri: requestUri,
+      name: cieData.name,
+      surname: cieData.surname,
+      birth_date: cieData.birthDate,
+      fiscal_code: cieData.fiscalCode,
+    });
```
balanza, line 174: **question(privacy):** It's ok to pass user's personal data as plain query string? 

Theoretically speaking, it's safe to pass sensitive data as query string as long as the connection is over SSL (can we enforce that?). Even though, we still have the problem of urls being logged by web servers and proxies.  

Maybe it's not a concern for this PR but for the protocol itself, maybe it has been already discussed and it's fine. But I'd rather double-check the decision and make sure we have all the needed countermeasures in action.

src/utils/decoder.ts
```diff
@@ -0,0 +1,44 @@
...
+ *       value="eyJhbGciOiJSUz....."/>
+ *       </form>
+ *    </body>
+ *   </html>
+ */
+export const getJwtFromFormPost = async (
+  formData: string
+): Promise<{ jwt: string; decodedJwt: JWTDecodeResult }> => {
+  const formPostRegex = /<input(.|\n)*value\s*=\s*"((.|\n)*)"(.|\n)*>/gm;
+  const lineExpressionRegex = /\r\n|\n\r|\n|\r|\s+/g;
```
balanza, line 28: **suggestion:** It would be great to have a test covering the scenario of those characters being trimmed. ‏

src/pid/issuing.ts
grausof: Where is the sensitive data? The data that is passed now is identifying data only. However, this is a workaround while waiting for authentication with "Enter with CIE" to be implemented, where the identification data will be obtained directly from the service provider.

src/utils/decoder.ts
grausof: Unit test is [here](https://github.com/pagopa/io-react-native-wallet/pull/47/files#diff-43df75678a472f5960db435df13e70167d096dca18db5c9fe53f69a2d6e1033d). Do you think it needs to be grittier?

src/pid/issuing.ts
balanza: Ok my bad, I thought we were passing them here as well

src/utils/decoder.ts
balanza: I don't know, maybe one more example showing the need for the replace? Just to stress the utility a bit on edge cases.

src/pid/issuing.ts
balanza: Although it's perfectly fair for the current status of the project, given the importance and visibility of the initiative I would write an extensive comment on how this is a temporary workaround.
Hopefully, this source will drive the interest of the citizens, and someone may make the same assumption I did. 

For example
```ts
// User's personal data is not supposed to transit in this flow,
// but to be provided to the PID issuer directly by its chosen authentication method (CIE).
// Being the project in an initial phase, and being we were still unable to fully comply with authentication,
// we temporarily provide data from the App's logged user.
```

balanza (approved):

src/utils/decoder.ts
grausof: the replace of new lines is also tested in the unit test

hevelius (approved):
LGTM. It works as expected.

PR #50, opened 2023-10-05
feat: [SIW-323] Verify Wallet Instance Attestation on receving

balanza:
#### List of Changes

* verify wallet instance attestation jwt.

#### Motivation and Context

Before finalizing the Wallet Instance Attestation issuing process, we validate the just received attestation by verifying:
* its signature is compliant with what exposed by the Wallet Provider Entity Configuration
* the token has not expired
* the claims contain correct wallet provider data

#### Checklist:

### DIFF
--- a/src/wallet-instance-attestation/issuing.ts
+++ b/src/wallet-instance-attestation/issuing.ts
@@ -51,10 +51,17 @@ async function verifyWalletInstanceAttestation(
   wia: string,
   walletProviderEntityConfiguration: WalletProviderEntityConfiguration
 ): Promise<true> {
-  return verifyJwt(
-    wia,
-    walletProviderEntityConfiguration.payload.metadata.wallet_provider.jwks.keys
-  )
+  const {
+    payload: {
+      sub,
+      metadata: {
+        wallet_provider: {
+          jwks: { keys },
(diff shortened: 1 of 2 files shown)


src/wallet-instance-attestation/issuing.ts
```diff
@@ -38,6 +38,40 @@ async function getAttestationRequest(
...
 }
 
+/**
+ * Validate a Wallet Instance Attestation token.
+ * Either return true or throw an exception.
+ *
+ * @param wia Signed Wallet Instance Attestation token
+ * @param walletProviderEntityConfiguration Entity Configuration object for the issuing Wallet Provider
+ * @returns The token is valid
+ * @throws {WalletInstanceAttestationIssuingError} When the received token fails to validate. This can happen due to invalid signature, espired token or malformed JWT token.
```
hevelius, line 48:
```suggestion
 * @throws {WalletInstanceAttestationIssuingError} When the received token fails to validate. This can happen due to invalid signature, expired token or malformed JWT token.
```

hevelius (changes requested):

grausof (approved):
LGTM

PR #51, opened 2023-10-06
fix: [SIW-584] Fix DPoP endpoint

balanza:
#### List of Changes

* fix wrong endpoint
* refactor ephemeral key context

#### Motivation and Context

DPoP tokens must be bound to the request they come with. We were creating the DPoP token for a wrong endpoint

Example app

#### Checklist:


src/pid/issuing.ts
```diff
@@ -341,26 +335,26 @@ export const getCredential =
     { nonce, accessToken, clientId, walletProviderBaseUrl }: AuthorizationConf,
     pidProviderEntityConfiguration: CredentialIssuerEntityConfiguration
   ): Promise<PidResponse> => {
+    const credentialUrl =
+      pidProviderEntityConfiguration.payload.metadata.openid_credential_issuer
+        .credential_endpoint;
```
balanza, line 340: Just moved the declaration up

src/pid/issuing.ts
```diff
@@ -341,26 +335,26 @@ export const getCredential =
...
+
     const signedDPopForPid = await createDPopToken(
       {
         htm: "POST",
-        htu: pidProviderEntityConfiguration.payload.metadata
-          .openid_credential_issuer.token_endpoint,
+        htu: credentialUrl,
```
balanza, line 345: The actual fix

src/pid/issuing.ts
```diff
@@ -244,22 +243,17 @@ export const authorizeIssuing =
...
+          htm: "POST",
+          htu: tokenUrl,
+          jti: `${uuid.v4()}`,
+        },
+        ctx
+      )
     );
 
-    await deleteKey(keytag);
-
```
balanza, line 262: Moved the key-handling logic into `useEphemeralKey` utility

src/utils/crypto.ts
```diff
@@ -39,3 +45,28 @@ export const createCryptoContextFor = (keytag: string): CryptoContext => {
...
+export const useEphemeralKey = async <R>(
+  fn: (ephemeralContext: CryptoContext) => Promise<R>
+): Promise<R> => {
+  // Use an ephemeral key to be destroyed after use
+  const keytag = `ephemeral-${uuid.v4()}`;
+  await generate(keytag);
+  const ephemeralContext = createCryptoContextFor(keytag);
+  try {
+    const result = await fn(ephemeralContext);
+    await deleteKey(keytag);
```
gunzip, line 66: maybe we can put `deleteKey` into finally {} ?

balanza: I wasn't sure how `finally` works with async code, I suspected some quirks and I was too lazy to investigate 🙃.

It seems to work as expected. However, I fail to test it on a real device (so I just hope iOS/Android js engine likes it as well).

On the happy path, it works like charm https://github.com/pagopa/io-react-native-wallet/pull/51/commits/8c5dfe769953004e77df4f7b0ea44dc1a5416398

@LazyAfternoons @hevelius any thoughts?

src/utils/crypto.ts
```diff
@@ -39,3 +45,27 @@ export const createCryptoContextFor = (keytag: string): CryptoContext => {
...
+ * An ephemeral crypto context is a context which is bound to a key
+ * that is just created and is deleted after use.
+ *
+ * @param fn The procedure to be executed
+ * @returns The returned value of the input procedure.
+ */
+export const useEphemeralKey = async <R>(
+  fn: (ephemeralContext: CryptoContext) => Promise<R>
+): Promise<R> => {
+  // Use an ephemeral key to be destroyed after use
```
grausof, line 60: I'm not sure this is the right approach. You may also need to access the temporary public key later in your code (for example to check the holder binding with the response). It is correct to delete it from the TEE but you need to leave the value of the public key to be able to use it later.

balanza: As a user, you can get the public key as a result, if needed:
```ts
const [result, pubk] = await useEphemeralKey(ctx => {
  const publicKey = await ctx.getPublicKey();
  // your stuff
  return [value, publicKey];
});
```

grausof (approved):

src/utils/crypto.ts
```diff
@@ -62,11 +62,10 @@ export const useEphemeralKey = async <R>(
...
   try {
-    const result = await fn(ephemeralContext);
-    await deleteKey(keytag);
-    return result;
+    return fn(ephemeralContext);
   } catch (error) {
-    await deleteKey(keytag);
     throw error;
+  } finally {
+    await deleteKey(keytag);
```
grausof: I think you should leave `await`:
`return await fn(ephemeralContext);`
otherwise the `finally` is called before `fn` finishes.

PR #52, opened 2023-10-06
feat: [SIW-490] Trust chain renewal

balanza:
#### List of Changes

* add utilities to fetch signed Entity Statement and Entity Configuration
* add utility to renew trust chain
* add validation test on received credential

#### Motivation and Context

Trust chain shipped by credentials must be short-lived. Hence, we need a utility to obtain updated entity statements from an expired chain.

The proposed usage is
```ts
import {
  verifyTrustChain,
  getTrustAnchorEntityConfiguration,
} from "@pagopa/io-react-native-wallet";

const trustChain = ["ejJ0eX...", "eyG5eX...", "erU9eX..."];
const trustChainEC = await getTrustAnchorEntityConfiguration(
  "https://trust-anchor.example"
);

// Validate a given trust chain offline (no renewal on failures)
await verifyTrustChain(trustChainEC, trustChain);

// Validate a given trust chain online (try to renew on failures)
await verifyTrustChain(trustChainEC, trustChain, { renewOnFail: true });
```

#### Checklist:


example/src/scenarios/get-pid.tsx
```diff
@@ -3,12 +3,31 @@ import {
...
+    trustAnchorBaseUrl
+  );
+
+  // test verify trust chain
+  await verifyTrustChain(trustAnchorEntity, trustChain);
+
+  // test renew of the trust chain
+  // (just for test, needed only when input chain fails to validate)
+  const renewedChain = await renewTrustChain(trustChain);
+  await verifyTrustChain(trustAnchorEntity, renewedChain);
```
grausof, line 29: What do you think about adding check here if the trust chain is expired? Or maybe add a utility function so that it can be used easily?

src/trust/chain.ts
```diff
@@ -107,3 +108,40 @@ export async function verifyTrustChain(
...
+      // Decode each item to determine its shape
+      .map(decode)
+      .map(
+        (e) =>
+          [
+            EntityStatement.safeParse(e),
+            EntityConfiguration.safeParse(e),
+          ] as const
+      )
+      // fetch the element accorfing to its shape
```
grausof, line 135:
```suggestion
      // fetch the element according to its shape
```

src/trust/chain.ts
```diff
@@ -107,3 +108,40 @@ export async function verifyTrustChain(
...
+      .map(([es, ec], i) =>
+        ec.success
+          ? getSignedEntityConfiguration(ec.data.payload.iss, { appFetch })
+          : es.success
+          ? getSignedEntityStatement(es.data.payload.iss, es.data.payload.sub, {
+              appFetch,
+            })
+          : // if the element fail to parse in both EntityStatement and EntityConfiguration, raise an error
+            Promise.reject(`Failed to parse element #${i} of the trust chain`)
```
grausof, line 144: Here you could use the `Error` object common to the entire library.

grausof (changes requested):
Some small changes. The rest is excellent for me!

example/src/scenarios/get-pid.tsx
balanza: You mean something like
```ts
async function trust(trustAnchorEntity, chain){
  try {
    return verifyTrustChain(trustAnchorEntity, chain) 
  } catch(ex){
    const renewedChain = await renewTrustChain(chain);
    return verifyTrustChain(trustAnchorEntity, renewedChain);
  }
}
```
Sure It will be useful, we can do it.

I'm not sure if we want to discriminate whether the trust chain verification fails for expiration or something else. I thought we could just try again with a renewed chain on every failure.

grausof: Yep. I would only throw an exception if renewal fails or post renewal verification fails.

balanza: It will, as both `renewTrustChain` and `verifyTrustChain` may throw an exception. Do you mean a more descriptive error? Sounds resonable.

src/trust/chain.ts
balanza: Fixed in https://github.com/pagopa/io-react-native-wallet/pull/52/commits/be1ddbe72fcc8c7015a832063d52a608399d312f

example/src/scenarios/get-pid.tsx
balanza: I proposed another interface [here](https://github.com/pagopa/io-react-native-wallet/pull/52/commits/be1ddbe72fcc8c7015a832063d52a608399d312f)

grausof (approved):
Great work 🚀

PR #53, opened 2023-10-09
Fix RP entity configuration

grausof:
Fix the entity configuration type for the relying party

#### List of Changes
Change `metadata->wallet_relying_party->jwks` to `metadata->wallet_relying_party->jwks->keys`


hevelius (approved):
LGTM

PR #57, opened 2023-10-10
Small syntax fix

balanza:
It was so easy I'm ashamed I didn't think about it

image


grausof (approved):
1ffff6d108ebf73c5d130ca7f4ead52c

PR #62, opened 2023-10-23
feat!: [SIW-453] Credential Issuing flow

balanza:
This PR adds a general flow for credential issuing. 
In order to do it, it introduces a **huge** refactor of both issuing and presentation flow. Below is an overview of changes to help readers to navigate through file diffs.

### List of Changes

**Package API**
The API exposed by the package has been reorganized using the following namespaces:
* `Credential.Issuance` contains the functions and types that map the steps of the issuing flow
* `Credential.Presentation` contains the functions and types that map the steps of the presentation flow
* `Trust` contains the functions and types that interact with the trusted entities

```ts
import { Credential, Trust } from "@pagopa/io-react-native-wallet"

Credential.Issuance.obtainCredential(...)
Credential.Presentation.getRequestObject(...)
Trust.getEntityStatement(...)
```
see `src/index.ts`

**Removed modules**
The modules `src/pid` and `src/rp` have been removed. Their implementations have been moved into `credential` as they are now generalized. `src/pid/sw-jwt` is still there as it implements the PID model.

**Code organization**
Being that both Issuance and Presentation flows are processes that concatenate input and output from a series of functions,
* One file for every step exposing the type that defines the input and output of the step, and its implementation
  * see files in `src/credential/issuance` and `src/credential/presentation`
* Each file is named with a number so that they are listed in the same order as they appear in the flow* 
* Steps that do not have a concrete implementation still have a type definition, so that the user application can develop it's own implementation; the rationale is that some operations require User interactions through UI elements that are out of scope for this package
  * see  `src/credential/issuance/01-start-flow.ts` as an actual example
* Function signatures are designed with the same shape:
  * they take as input the output of previous steps and the user-provided data
  * the last argument is a `context` object that maps what the step need from the environment (keys, attestations, configurations, etc)
  * they always return `Promise<Record<string, T>>`, even when only one value is returned; the rationale is to have a more eloquent return type
  * they throw in the operation fails

**utils: `Out<FN>`**
Just an alias for `Awaited<ReturnType<FN>>`, useful to link one function's output as another's input.
Rationale: make connections through steps more evident.

```ts
import { type Out } from "src/utils/misc"

declare function fn1(): Promise<{ foo: string }>
declare functon fn2(foo: Out<typeof fn1>["foo"]): Promise<void>
```

**utils: `hasStatus(number)`**
Simple helper that takes a `Response` object and throws an exception if it's not the expected status.
Rationale: code reuse and standardization

```ts
// before
const response = await appFetch("<url>")

if (response.status === 200) {
  const payload = await response.json();
  return { foo: payload.foo, bar: payload.bar };
}
throw new CustomError(
  `Unable to do something. Response code: ${await response.text()}`
);

// after
import { hasStatus } from "src/utils/misc"

return appFetch("<url>")
  .then(hasStatus(200)) // <-- a standard Http Error will be thrown
  .then(res => res.json())
  .then(payload => ({ foo: payload.foo, bar: payload.bar });
```

**Things fixed on the way**
* Add `credential_definition` for the supported credential metadata in `src/trust/types.ts`
* Add optional `wallet_relying_party` metadata for Credential issuer in `src/trust/types.ts`
* Make credential `type` generic in SD-JWT schema in `src/sd-jwt/types.ts`

**Workarounds**
* ~When obtaining Credential, before returning the received token we validate it. The signature validation fails for now, it's maybe an error in the Issuer? For now, we just log the error and skip the validation (see https://github.com/pagopa/io-react-native-wallet/blob/3772a5210e563cfdea44f7c3fc3dd27fd33b4e37/src/credential/issuance/06-obtain-credential.ts#L115)~ fixed
* Ignore missing `response_code` in https://github.com/pagopa/io-react-native-wallet/blob/e21362908b7b4ba31ef28dded4b55b8116242526/src/credential/presentation/04-send-authorization-response.ts#L18
* Normalize PID Issuer response with stub data so that is compliant to Credential issuance specification in https://github.com/pagopa/io-react-native-wallet/blob/dd1048632bac5348cf31c77e8962b0152e50c6ff/src/credential/issuance/03-start-user-authorization.ts#L39

### How Has This Been Tested?
Example app

### Checklist:

----
Rewrite of #54


balanza:
Open points:
* are we fine with step naming?
* are we fine with the proposed API?

src/credential/issuance/04-complete-user-authorization.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * A dummy implementation of CompleteUserAuthorization that uses static values.
+ * Used to replace unimplemented specifications by the Issuer
+ */
+export const completeUserAuthorizationNoOp: CompleteUserAuthorization = async (
+  _,
+  __
+) => {
+  return { code: "static_code" };
+};
```
grausof, line 28: Why is this specification not yet implemented? Could you detail it better? Maybe we also add a TODO with a reference task

src/credential/issuance/06-obtain-credential.ts
```diff
@@ -0,0 +1,141 @@
...
+  credential: z.string(),
+  format: z.literal("vc+sd-jwt"),
+  /* YAGNI
+  Ideally, we may request more than one credential after a single authorization phase
+  Each new call to the credential endpoint must carry a different nonce value,
+  otherwise it would be rejected.
+  So every response has a new nonce value to eventually be used in the next request.
+  It's not implementd so far
+  c_nonce: z.string(),
+  c_nonce_expires_in: z.number().positive(),
```
grausof, line 26: TODO?

src/credential/issuance/06-obtain-credential.ts
```diff
@@ -0,0 +1,141 @@
...
+  // validate the received credential signature
+  // is correct and refers to the public keys of the issuer
+  const decodedCredential = await verifyJwt(
+    credential,
+    issuerConf.openid_credential_issuer.jwks.keys
+  ).catch((_) => {
+    // FIXME: [SIW-629] signature verification failures is ignored
+    // as issuer does not implement this feature correctly.
+    // Just logging for now.
+    console.error(_);
```
grausof, line 122: ⚠️ I think this is not acceptable. The problem must be fixed by the issuer if it exists

src/credential/presentation/03-get-request-object.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+    if (!pubKey) {
+      throw new NoSuitableKeysFoundInEntityConfiguration(
+        "Request Object signature verification"
+      );
+    }
+    await verify(responseEncodedJwt, pubKey);
+  }
+
+  // parse request object it has the expected shape by specification
```
grausof, line 79:
```suggestion
  // Ensure that the request object conforms to the expected specification.
```

src/credential/presentation/04-send-authorization-response.ts
```diff
@@ -0,0 +1,175 @@
...
+    .string() /**
+      FIXME: [SIW-627] we expect this value from every RP implementation
+      Actually some RP does not return the value
+      We make it optional to not break the flow.
+    */
+    .optional(),
+});
+
+/**
+ * Select a RSA public key from those provided by the RP to encrypt.
```
grausof, line 26:
```suggestion
 * Choose an RSA public key from those offered by the RP for encryption.
```

src/credential/presentation/04-send-authorization-response.ts
```diff
@@ -0,0 +1,175 @@
...
+  );
+};
+
+/**
+ * Prepare the Verified Presentation token for a received request object in the context of an authorization request flow.
+ * The presentation is prepared by disclosing data from provided credentials, according to requested claims
+ * Each Verified Credential come along with the claims the user accepts to disclose from it.
+ *
+ * @todo accept more than a Verified Credential
+ */
```
grausof, line 55:
```suggestion
/**
 * Generate a Verified Presentation token for a received request object within the context of an authorization request flow.
 * The presentation is created by revealing data from the provided credentials based on the requested claims.
 * Each Verified Credential is accompanied by the claims that the user consents to disclose from it.
 *
 * @todo: Allow for handling more than one Verified Credential.
 */
```

src/credential/presentation/04-send-authorization-response.ts
```diff
@@ -0,0 +1,175 @@
...
+export const sendAuthorizationResponse: SendAuthorizationResponse = async (
+  requestObject,
+  rpConf,
+  presentation,
+  { appFetch = fetch, walletInstanceAttestation }
+): Promise<AuthorizationResponse> => {
+  // the request is an unsigned jws without iss, aud, exp
+  // https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-signed-and-encrypted-respon
+  const jwk = chooseRSAPublicKeyToEncrypt(rpConf);
```
grausof, line 141: I'll put it here for future reference. Better to use speaking variable names
```suggestion
  const rsaPublicJwk = chooseRSAPublicKeyToEncrypt(rpConf);
```

src/credential/presentation/04-send-authorization-response.ts
```diff
@@ -0,0 +1,175 @@
...
+    appFetch?: GlobalFetch["fetch"];
+  }
+) => Promise<AuthorizationResponse>;
+
+/**
+ * Compose and send an Authorization Response in the context of an authorization request flow.
+ *
+ * @todo MUST add presentation_submission
+ *
+ */
```
grausof, line 120: Is there a reference for this task?

src/credential/issuance/06-obtain-credential.ts
balanza: Fortunately, it was just a mistake of us. Fixed in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/c5c8c37d35381f3b7d83186018a04cfc4d8f0ff0

src/credential/issuance/06-obtain-credential.ts
balanza: We don't really need commented code. Removed in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/32063c845f6fbc0b5410eda0904733319569ca16

src/credential/issuance/04-complete-user-authorization.ts
balanza: I agree. I removed this function and implemented _no-op_ in the example app in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/eb3a27676c9120b747d2256766b8479fb2a2c23d

src/utils/par.ts
```diff
@@ -0,0 +1,131 @@
...
+    walletProviderBaseUrl: string,
+    parEndpoint: string,
+    walletInstanceAttestation: string,
+    authorizationDetails: AuthorizationDetails,
+    assertionType: string
+  ): Promise<string> => {
+    // Calculate the thumbprint of the public key of the Wallet Instance Attestation.
+    // The PAR request token is signed used the Wallet Instance Attestation key.
+    // The signature can be verified by reading the public key from the key set shippet with the it will ship the Wallet Instance Attestation;
+    //  key is matched by its kid, which is supposed to be the thumbprint of its public key.
```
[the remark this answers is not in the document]
balanza: I rearranged the code in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/454042f6f2351d5716c7a0c2c4e2377838513c98, I guess it makes more sense now

src/credential/presentation/04-send-authorization-response.ts
balanza: It was actually a wring comment, removed in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/42caeb6fcedc07aa5990371f87c80d3efdb3c8ea

src/trust/types.ts
```diff
@@ -113,6 +139,24 @@ export const CredentialIssuerEntityConfiguration = BaseEntityConfiguration.and(
           credentials_supported: z.array(SupportedCredentialMetadata),
           jwks: z.object({ keys: z.array(JWK) }),
         }),
+        wallet_relying_party: RelyingPartyMetadata.optional() /*
+        Should it be mandatory? Apart from PID issuer, a credential issuer 
+        must operate as relying party to accept user authorization through presentation
+        */,
```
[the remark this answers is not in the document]
balanza: done in https://github.com/pagopa/io-react-native-wallet/pull/62/commits/42caeb6fcedc07aa5990371f87c80d3efdb3c8ea

balanza:
> Is it possible to investigate what is due to this?

I couldn't actually replicate the issue. I'm executing the app on a physical Android device (OnePlus 6T)

grausof:
> > Is it possible to investigate what is due to this?
> 
> I couldn't actually replicate the issue. I'm executing the app on a physical Android device (OnePlus 6T)

After the latest commits it seems that the problem has disappeared!

grausof (approved):
Really great work 🚀 
P.S. Given the size of the PR I would also wait for another approve from @hevelius before the merge.
☝ **Remember to change the package version!**

hevelius (approved):
LGTM. Tested on simulator without errors. Great job!

PR #63, opened 2023-10-26
fix!: [SIW-631] `StartFlow` to be syncronous

balanza:
#### List of Changes

* `StartFlow` as sync function
* `Out` to handle both sync and async functions

#### Motivation and Context

`StartFlow` is a generic function signature that maps any touchpoint the User can use to start the Issuance and Presentation processes. It was introduced in #62.
Touchpoint may be:
* QR code scanning (cross-device)
* Select an element from a list (same device)
* URL decode (same device)

We found out it's better to keep it synchronous

Type system check

#### Checklist:


grausof (approved):

hevelius (approved):
LGTM

PR #69, opened 2023-12-06
feat: [SIW-678] Support issuing credential in `mdoc-cbor` format

balanza:
#### List of Changes

* Allow mdoc-cbor format
* Allow multiple credential requests with a single authorization profile

#### Motivation and Context

As per ARF, credentials are served in SD-JWT or mDoc-CBOR format. 
Moreover, for many credentials both formats are provided thus we enable the user to obtain both at once.

example app

#### Checklist:

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -55,6 +55,10 @@ export default function App() {
           title="Get Credential"
           scenario={scenarios.getCredential}
         />
+        <TestScenario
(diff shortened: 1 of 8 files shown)


grausof (approved):
Great work 🚀

balanza:
> What do you think about installing [changesets](https://github.com/changesets/changesets)?

not today

PR #71, opened 2023-12-13
fix: [SIW-762] Credential parsing with undefined attributes

LazyAfternoons:
Co-authored by @balanza 

#### List of Changes
Converts the list of disclosures key-value pairs into an object before constructing the list of attributes that are not defined in the issuer configuration.

#### Motivation and Context
This change fixes an issue which occurs when the credential has disclosures not defined in the issuer configuration and the `ignoreMissingAttributes` is set to `true`. 
The issue is caused by a missing conversion of the disclosures key-value pairs into an object which causes a duplication of defined disclosures in the undefined list. 
This leads to an overwrite of the defined disclosure when the two lists are merged in the function return.

By testing the `mDL` issuance (SD-JTW). The verified and parsed credential should have each disclosures with a locale.

#### Checklist:

### DIFF
--- a/src/credential/issuance/07-verify-and-parse-credential.ts
+++ b/src/credential/issuance/07-verify-and-parse-credential.ts
@@ -89,45 +89,49 @@ const parseCredentialSdJwt = (
 
   // attributes that are defined in the issuer configuration
   // and are present in the disclosure set
-  const definedValues = attrDefinitions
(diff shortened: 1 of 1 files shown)


hevelius (approved):
LGTM. I have not clear how to test this PR and if enough to try the example app but it works correctly.

PR #74, opened 2024-04-22
chore: Add production section to example app

grausof:
Adds a new section to the sample app by splitting the screen between PoC and Production.

Screenshot 2024-04-22 at 15 06 30


example/src/App.tsx
```diff
@@ -38,38 +42,83 @@ export default function App() {
...
+              if (route.name === "PoC") {
+                icon = "🧪";
+              } else if (route.name === "Production") {
+                icon = "🚀";
+              }
+              return <Text>{icon}</Text>;
+            },
+            tabBarActiveTintColor: "blue",
+            tabBarInactiveTintColor: "gray",
+          })}
```
LazyAfternoons, line 60: Small suggestion here, we could extract this function like this: 
`const bottomBar = (
    route: RouteProp<RootStackParamList, keyof RootStackParamList>
  ) => {
    let icon;
    if (route.name === "PoC") {
      icon = "🧪";
    } else if (route.name === "Production") {
      icon = "🚀";
    }
    return <Text>{icon}</Text>;
  };`
  
  And use it like: 
`tabBarIcon: () => bottomBar(route),`

example/src/App.tsx
LazyAfternoons, line 1: We could use `SafeAreaView` exported by `react-native-safe-area-context` which you already added as dependency as it also works on Android, rather than the standard export from `react-native` which only works on iOS. 

I would also remove the `marginHorizontal` from the `container` style as it adds it also around the tab bar. Maybe we could add it in an inner view only for the content of the screen, leaving the tab bar out. 

Small changes but let me know what you think.

LazyAfternoons (changes requested):

example/src/App.tsx
grausof: That's right, you're a React ninja 🥷.
I added a `SafeAreaProvider` and a `SafeAreaView` from `react-native-safe-area-context`. 
Let me know if it is okay!

hevelius (changes requested):

hevelius (approved):
LGTM. @LazyAfternoons please check if your conversations can be resolved.

LazyAfternoons (commented):
LGTM

PR #81, opened 2024-05-17
feat: [SIW-1076] Add create wallet instance function

grausof:
Adds the function for creating the Wallet Instance. Has been added:
- The Wallet Provider's openapi
- Dynamic type generation from openapi in zod via [typed-openapi](https://github.com/astahmer/typed-openapi)
- Added Wallet Provider client
- Introduced integrityContext

#### Motivation and Context
The Wallet Solution needs to register its wallet instance on the Wallet Provider's backend, uniquely identified by a securely stored hardware key.

Tested locally with the sample app on iOS only. Some additional testing needed on Android.


openapi/wallet-provider.yaml
gispada, line 1: Do we really need to hardcode the OpenAPI definition here? `typed-openapi` seems to work with a remote url, can't we generate the client like in `io-app`, pointing to the raw github file?

src/client/generated/wallet-provider.ts
gispada, line 1: If we want to use a generated client, I would not commit it.

src/client/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  ProblemDetail,
+  createApiClient as createWalletProviderApiClient,
+} from "./generated/wallet-provider";
+import { ApiClient as WalletProviderApiClient } from "./generated/wallet-provider";
+
+export type WalletProviderClient = WalletProviderApiClient;
+
+const validateResponse = async (response: Response) => {
+  if (!response.ok) {
+    var problemDetail: ProblemDetail = {};
```
gispada, line 12: Why `var`?

package.json
```diff
@@ -34,7 +34,8 @@
     "release": "release-it",
     "example": "yarn --cwd example",
     "bootstrap": "yarn example && yarn install",
-    "code-review": "yarn lint && yarn typecheck && yarn test"
+    "code-review": "yarn lint && yarn typecheck && yarn test",
+    "generate-wallet-provider-client": "cd openapi && typed-openapi wallet-provider.yaml -o ../src/clients/generated/wallet-provider.ts -r zod && cd ../"
```
silvicir, line 38: `"typed-openapi openapi/wallet-provider.yaml -o src/client/generated/wallet-provider.ts -r zod"` ?

src/client/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      problemDetail = ProblemDetail.parse(await response.json());
+    } catch (_) {
```
silvicir, line 15:
```suggestion
    } catch {
```

src/client/index.ts
```diff
@@ -0,0 +1,55 @@
...
+        ? problemDetail.title
+        : "Invalid response from Wallet Provider",
+      problemDetail.type,
+      problemDetail.detail
+    );
+  }
+  return response;
+};
+
+export function getWalletProviderClient(context: {
```
silvicir, line 33: I would use the same method for defining functions (all through function declarations or through function expressions)

src/client/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  walletProviderBaseUrl: string;
+  appFetch?: GlobalFetch["fetch"];
+}) {
+  const { walletProviderBaseUrl, appFetch = fetch } = context;
+
+  return createWalletProviderApiClient(
```
silvicir, line 39: shouldn't we handle the case where the `fetch` fails?

package.json
gunzip: note that `typed-openapi` types are pretty naive (it doesn't support dates, pattern strings, binary, ...). Moreover the community is not that big and latest commit was 3 moths ago. I don't know where this code will be used, but you may consider to use [openapi-zod-client](https://github.com/astahmer/openapi-zod-client) instead, which is pretty similar but has a better community, looks maintained and supports complex types.

grausof: Unfortunately `openapi-zod-client `uses `axios` internally via the `zodios` client API and it is not possible to change the fetch function programmatically. It is not possible to use it here as the react-native based library is used by app IO that **do not have a node instance** to make `zodios` work and it is also necessary that the user of this library can declare the fetch function via a context to use. 

To overcome these limitations `openapi-zod-client` itself recommends using `typed-openapi`.

Since the open API of the entire wallet is always very simple, these are strings containing JWT encoded in base64, I believe that this library is sufficient. Do you have any other generator in mind that could be right for us?

gunzip: I've assessed sever generators and found no one that totally suits our needs. Suprisingly [our homemade generator](https://github.com/pagopa/openapi-codegen-ts), the one that io-app already uses for all APIs, is well positioned, but requires to relax the `zod` usage constraint since it's `io-ts` based.

grausof: 😞 `io-react-native-wallet` does not use **fp-ts** in favor of pure **typescript**. I would say to leave this and then in the future evaluate whether it is worth transforming our homemade generator for zod too.

openapi/wallet-provider.yaml
grausof: Yes correct, but for now we do not yet have these specifications published on io-backend as the proxy has not yet been developed, so for now a copy is left here

src/client/generated/wallet-provider.ts
grausof: Right. Leaving the file generated for this PR in draft is intended to give visibility during the review of how the file is generated to also allow the generator used to be evaluated. I will delete it before merging the PR.

src/client/index.ts
grausof: Right, thanks

src/client/index.ts
grausof: True, I hadn't thought of that. Then I would say not to catch

LazyAfternoons: We might refine this by exposing our own subset of errors including the error code. This could be especially useful to distinguish `401` returned from our backend or from our partners which might have a different meaning (implying that there are flows with mixed calls). That's just an idea though, it might unwrap differently later on.

grausof: The error codes, however, do not imply a fetch throw but simply a response with a different status code. For now the `validateResponse` in case of error throws a classic `WalletProviderResponseError` with the details of the error. Maybe with a future PR this could be specialized. 
For now I would say to leave it like this since the error is telling anyway.

grausof: I added more details about the status code error on `validateResponse`

src/client/generated/.keep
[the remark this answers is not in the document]
grausof: Otherwise the `typed-openapi` generator fails to find the `src/client/generated/` folder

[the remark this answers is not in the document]
grausof: In my opinion it is better to use an ad hoc folder to better manage the `.gitignore` or `.eslintignore` (otherwise you would have to manually insert all the generated files) and to have a single place where all the auto-generated files are.

An alternative would be to add a `mkdir` to the `generate` command. 

What do you think?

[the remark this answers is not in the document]
silvicir: got it 👍 as you want, I'm fine with both

silvicir (approved):

PR #84, opened 2024-05-23
chore: [SIW-1160] Add Android example app

LazyAfternoons:
This PR adds the android example app for the `createWalletInstance` scenario.

#### List of Changes

- Use `Platform.select` to differentiate between the iOS and Android flow when generating the hardware key. The main difference is that on Android we generate the hardware-backed key via `io-react-native-crypto`.

Follow these steps to test the flow: 
- Run [io-func-wallet-solution-backend](https://github.com/pagopa/io-wallet/tree/master/apps/io-func-wallet-solution-backend). I can provide the required `local.settings.json` in private if necessary;
- Run the proxy server under `example/dev-proxy`. I can provide the required `.env` in private if necessary. Make sure that `API_HOST` corresponds to the address:port combination of the function started in the previous step (defaults to `http://localhost:7071`);
- Run the example app. I can provide the required `.env` in private if necessary. Make sure that `WALLET_PROVIDER_BASE_URL` corresponds to the address:port combinato of the dev-proxy started in the previous step (defaults to your localhost address).

It has been tested on different Android devices (9,13,14).

#### Checklist:

### DIFF
--- a/example/src/scenarios/create-wallet-instance.tsx
+++ b/example/src/scenarios/create-wallet-instance.tsx
@@ -7,62 +7,21 @@ import {
   generateHardwareKey,
   getAttestation,
   isAttestationServiceAvailable,
-  isPlayServicesAvailable,
 } from "@pagopa/io-react-native-integrity";
 
 import { WALLET_PROVIDER_BASE_URL } from "@env";
-import { generate, type CryptoError } from "@pagopa/io-react-native-crypto";
+import { generate } from "@pagopa/io-react-native-crypto";
 import { Platform } from "react-native";
 
 const walletProviderBaseUrl = WALLET_PROVIDER_BASE_URL;
 
-export default async () =>
-  Platform.select({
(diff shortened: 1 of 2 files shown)


example/src/scenarios/create-wallet-instance.tsx
```diff
@@ -39,3 +40,29 @@ export default async () => {
...
+  return await generateHardwareKey();
+};
+
+/**
+ * Generates the hardware backed key for Android.
+ * It differents from iOS because the key is generated via `io-react-native-crypto` instead of `io-react-native-integrity`.
+ * @returns a promise that resolves with the key tag as string.
+ */
+const generateKeyAndroid = async () => {
+  const keyTag = `${Math.random()}`;
```
grausof, line 65: What do you think about using UUID v4 to make it similar to iOS?

grausof (approved):

PR #88, opened 2024-06-08
chore: [SIW-1208] Use base64url for iOS integrity keys

LazyAfternoons:
#### List of Changes

- Edit `generateIntegrityHarwareKeyTag` to use the function exposed `io-react-native-jwt` to convert the `base64` key returned from the native method to `base64url`;
- Edit `getHardwareSignatureWithAuthData` and `getAttestation` to convert the `base64url` key used in this package to `base64` which is what `io-react-native-integrity` uses;
- Defines `getAttestation` for Android as well due to different implementation which previously converged into the same native method.

#### Motivation and Context

Currently, `io-wallet` might return a `500` while creating a wallet instance on iOS due to the hardware key tag being encoded in base64 and thus having characters which can't be stored on the underlying Cosmos DB.
This PR converts the `base64` key tag to `base64url` in the `IntegrityContext`, thus making it safe to be stored on Cosmos DB. 
This doesn't happen on Android as the we can decide the name of the key tag used.

I belive this change belongs to this package rather than `io-react-native-integrity` which should be agnostic.

- Run an instance of `apps/io-wallet-user-func` from [io-wallet](https://github.com/pagopa/io-wallet) with an appropriate `local.settings.json` (I can provide it privately) by pointing at the branch used for [this PR](https://github.com/pagopa/io-wallet/pull/120);
- Run an instance of `dev-proxy` in this example folder by providing a `.env` file pointing at the running azure function of the previous point (`http://localhost:7071/api/v1` by default). I can provide the remaining variables privately);
- Run the example app on both Android and iOS by providing a `.env` file. The `WALLET_PROVIDER_BASE_URL` should point to the IP of your local machine URL where the `dev-proxy` is running (again, I can provide the remaining variables privately). Consequentially press every available button in the app and everything should work just fine.

#### Checklist:

### DIFF
--- a/example/src/scenarios/prepare-integrity-context.ts
+++ b/example/src/scenarios/prepare-integrity-context.ts
@@ -1,11 +1,11 @@
 import { type IntegrityContext } from "@pagopa/io-react-native-wallet";
 
 import { error, result } from "./types";
-import { getAttestation } from "@pagopa/io-react-native-integrity";
 import {
   ensureIntegrityServiceIsReady,
   generateIntegrityHarwareKeyTag,
   getHardwareSignatureWithAuthData,
+  getAttestation,
 } from "../utils/integrity/integrity";
 
 type IntegrityContextSetter = React.Dispatch<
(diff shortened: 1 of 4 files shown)


grausof (approved):
❤️

PR #90, opened 2024-06-18
feat: [SIW 1190]: handle errors to determine a Wallet Instance status

gispada:
#### List of Changes
- Added `WalletInstanceRevokedError`
- Added `WalletInstanceNotFoundError`
- Handle errors in `getAttestation`

#### Motivation and Context

When getting a Wallet Instance Attestation the server may respond with specific HTTP status codes that define the status of a Wallet Instance, i.e. revoked or not found.

The library now catches those status codes and maps them to more meaningful errors.

The example application was run with a proxy to intercept responses and change the HTTP status code.

### DIFF
--- a/example/src/scenarios/get-attestation.ts
+++ b/example/src/scenarios/get-attestation.ts
@@ -2,6 +2,7 @@ import { generate } from "@pagopa/io-react-native-crypto";
 import {
   WalletInstanceAttestation,
   createCryptoContextFor,
+  Errors,
   type IntegrityContext,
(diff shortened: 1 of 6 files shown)


LazyAfternoons (commented):
Question: should we keep a generic `WalletInstanceAttestationIssuingError` to wrap a possible exception occurring which doesn't fall in the two case covered by the specific exceptions you defined?

LazyAfternoons (approved):
LGTM

PR #91, opened 2024-06-19
feat: [SIW-1255] Add issuer authentication

grausof:
It introduces a utility within the example app to generate a `fetch` function that adds the **authentication token** read by the `env` to the **headers** of the request to the issuer.

### DIFF
--- a/example/src/utils/fetch.ts
+++ b/example/src/utils/fetch.ts
@@ -1,9 +1,8 @@
 import { ISSUER_AUTH_TOKEN, ISSUER_BASE_URL } from "@env";
 
-function addAuthHeaders(
-  options: RequestInit,
-  authHeaders: Record<string, string>
-) {
+type AuthHeaders = Record<string, string>;
+
+function addAuthHeaders(options: RequestInit, authHeaders: AuthHeaders) {
   return {
     ...options,
     headers: {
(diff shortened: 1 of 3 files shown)


example/src/utils/fetch.ts
```diff
@@ -0,0 +1,29 @@
...
+  const requestUrl =
+    typeof request === "string" ? new URL(request) : new URL(request.url);
+  let authHeaders: Record<string, string> = {};
+
+  // Add the authentication header only if I am contacting the issuer URL
+  if (requestUrl.origin === new URL(ISSUER_BASE_URL).origin) {
+    authHeaders = {
+      Authorization: `${ISSUER_AUTH_TOKEN}`,
+    };
+  }
```
LazyAfternoons, line 26:
```suggestion
  const authHeaders = requestUrl.origin === issuerBaseUrl.origin 
    ? { Authorization: `Bearer ${issuerAuthToken}` } 
    : {};
```
Does it make sense to you?

grausof: Yes, much better. I'm only eliminating Bearer because they told us he won't be there 🫢

LazyAfternoons (approved):
LGTM

PR #92, opened 2024-06-21
fix: [SIW-1273] Fix dev-proxy and rename env variables

grausof:
As it is structured, the dev-proxy only works by proxing the requests that arrive on the first path (e.g. `/path`) but does not manage multipaths (e.g. `/api/v1/path`). Since our simulated backend will have to be called on a different base path (`/api/v1/wallet`) this **basepath** has also been added to the dev server.

Furthermore, the names of the environment variables have been changed in accordance with what has already been defined in other PRs.


LazyAfternoons (approved):
LGTM

PR #94, opened 2024-06-24
fix: [SIW-1279] Alignment with new technical rules

grausof:
Alignment with new technical rules based on ongoing integration tests


LazyAfternoons (approved):
LGTM

PR #96, opened 2024-07-03
feat: [SIW-986,SIW-1309] Integrate CieID login

grausof:
This PR introduces the use of **CieId** as a login mechanism both via **L2 browser and via app**.

Previously, `authorizationContext` was used which brought a `web-view` into the app via the [@pagopa/io-react-native-login-utils ](https://github.com/pagopa/io-react-native-login-utils)library for login with **SPID** (SAML2).

Since **CieID** needs to open the external browser to complete the authentication process, in the case of **CieId** `authorizationContext` **must no longer be used** but the **external browser must be opened** directly.

So the behavior of `startCredentialIssuance` varies depending on whether the optional `authorizationContext` parameter is passed or not. 

⚠️ Warning: for **CieID** it must be `undefined`.

It is therefore **MANDATORY** to have the app that will use this library **register a deeplink or universallink** that corresponds to the redirectUri. In the example app the custom url `iowalletexample://` is registered


src/credential/issuance/03-start-credential-issuance.ts
```diff
@@ -323,25 +316,49 @@ export const startCredentialIssuance: StartCredentialIssuance = async (
...
+      }
     });
 
+    Linking.openURL(authUrl);
+
+    /*
+     * Waits for 120 seconds for the identificationRedirectUrl variable to be set
+     * by the custom url handler. If the timeout is exceeded, throw an exception
+     */
+    await until(() => authRedirectUrl !== undefined, 120);
```
LazyAfternoons, line 355: Should we await both of them with a `Promise.all` which rejects as soon as the first one rejects?

grausof: Thanks, fixed in [d5859f7](https://github.com/pagopa/io-react-native-wallet/pull/96/commits/d5859f7f7679a5651429918b040d8658196ae43e)

src/utils/misc.ts
```diff
@@ -35,3 +35,36 @@ export const generateRandomAlphaNumericString = (size: number) =>
...
+ *
+ * @param conditionFunction - A function that returns a boolean value.
+ *                            The promise resolves when this function returns true.
+ * @param timeout - An optional timeout in seconds. The promise is rejected if the
+ *                  condition function does not return true within this time.
+ * @returns A promise that resolves once the conditionFunction returns true or rejects if timed out.
+ */
+export const until = (
+  conditionFunction: () => boolean,
+  timeout?: number
```
gunzip, line 52:
```suggestion
  timeoutSeconds?: number
```

Is this in seconds as stated  in the jsdoc above? It looks like it gets multiplied by 1000 below

grausof: Fixed, thanks!

hevelius (approved):

PR #97, opened 2024-07-03
fix: [SIW-1311] Fix `birth_date` field

grausof:
Change the PID field from `birthdate` to `birth_date` by modifying the unit tests.


LazyAfternoons (approved):
LGTM

PR #101, opened 2024-07-12
feat: [SIW-1120] Add credential issuance (MDL)

LazyAfternoons:
> [!WARNING]  
> `claims: ClaimsMetadata.optional()` is set to optional due to an issue with the issuer EC
> The MDL currently cannot be obtained due to an issue with the credential source endpoint

#### List of Changes
- Update `.env.example` to include the EAA issuer link and update `config.ts` to use it;
- Add `get-crendential.ts` scenario which implements the issuance flow as an example;
- Update `get-pid.ts` to accept a `React.SetStateAction` which is called at the end of the flow to set the eID in the app state;
- Update `App.tsx` to include a test scenario to obtain a EAA credential, the MDL. Also add a state which holds the `eID` and its crypto context in order to obtain such credential;
- Update documentation for different scenarios based on the changes included in this PR.
- Update `04-complete-user-authorization` which is the main branching between the eID issuance flow and a generic credential. It includes two different new functions. Note that these two function might be merged, however after `getRequestedCredentialToBePresented` the user must confirm the willing to share the requested credentials in order to obtain the new one:
   - `getRequestedCredentialToBePresented` which must be called after `startUserAuthorization` and returns the request object containing the credentials required to obtain the EAA (currently eID and wia); 
  - `completeUserAuthorizationWithFormPostJwtMode` which completes the user authorisation and presents the requested credential and parses the authroisation server response as `form_post.jwt`. A new `zod` shape has been introduced, `ResponseUriResultShape`, for this purpose.
- Update  `getJwtFromFormPost` function to parse the `form_post.jwt` HTML page; 
- Update `07-verify-and-parse-credential` by reintroducing the `ignoreMissingAttributes` which comes in handy during tests and by handling the case where the claims metadata are undefined for a certain credentials.

#### Motivation and Context
This PR introduces the credential issuing flow for a generic credential. Currently only MDL is supported, thus we only have such scenario in the example app. 

Test every single scenario until you get a credential.

N/A

#### Checklist:


grausof (approved):
Great work 🎉
It just needs to be aligned to master

hevelius (approved):

PR #102, opened 2024-07-12
feat: [SIW-1293] Integrate eID issuance with CIE+PIN L3

grausof:
### Introduction
This PR integrates obtaining the eID via the **CIE+PIN L3** SDK ☢
In particular, the following changes have been made:
- The minimum **SDK** version of both **iOS** and **Android** has been changed to be compatible with the `@pagopa/react-native-cie` package
- For **iOS**, NFC permissions have been added to the sample app's `Info.plist` file
- For **Android**, manual linking of the` react-native-cie` library for `CiePackage` has been implemented in the sample app.  Also added dependency to `kotlin` and `okhttp3`.
- For the **Android** sample app, the ability to use **self signed certificates** has been added in order to use **ProxyMan**.
- The `Cie.WebViewComponent` was created which allows the reading of the CIE via NFC and the display of the **WebView** for user consent.
- In the example app, an **ad-hoc scenario** was created to test the entire flow by basing the **webview** on a **modal** that is opened or closed depending on the flow.

‼️ This component is based on the [react-native-webview](https://github.com/react-native-webview/react-native-webview) library since to use the L3 login it is necessary to **inject javascript code** inside the login page. For more details see the code documentation.

### How to use
To test this PR it is necessary to use a **physical device** and launch the `yarn cie-ios:prod` command before `yarn start`. If you are in a dev environment, however, you need to restore the packages with `yarn cie-ios:dev`

### Video
https://github.com/user-attachments/assets/4b621d13-f8d6-4664-8c81-e19bdbf000ff


hevelius (approved):
Static checks should be ok. Tested on iOS and Android physical devices. It works as expected. Tested with real CIE card.

LazyAfternoons (approved):
LGTM from a static perspective, I currently cannot test it as I don't have a CIE.

PR #105, opened 2024-07-17
feat: [SIW-1195] Add credential status implementation

LazyAfternoons:
This PR adds the credential status implementation. It currently works only for mDL.

#### List of Changes
- Add a `CredentialContext` which is set by the `getCredential` function in order to keep the obtained credential and its context in the app state. This will be refactored soon in another PR;
- Add `01-start-flow.ts` and `02-status-attestation` steps  in the status attestation flow. The first one simply specifies the issuer URL while the second one takes care of querying the `/status` endpoint for the given credential and parsing the result. It also maps the `404` status code to a custom exception which indicates that the credential has an invalid status attestation; 
- Add errors,types and utilities for this flow;
- Change the error thrown by `hasStatus` utility function from `IoWalletError` to `UnexpectedStatusCodeError`. This change is needed because `IoWalletError` doesn't carry the HTTP request status code, making it impossible to apply custom logics like the one needed for throwing a specific exception when the `/status` endpoint answers with a `404`.

#### Motivation and Context
The status attestation process enables the verification of credentials stored in a wallet, ensuring their validity given that they may be revoked at any time.

By using the example app in PRE env:
- Obtain a PID; 
- Obtain a mDL;
- Test the status attestation and it should return the credential attestation; 
- Revoke said credential (I can give the details in private on how to do this as it involves using secrets);
- Test the status attestation and it should rise the `StatusAttestationInvalid` exception;

#### Checklist:


src/credential/status/01-status-attestation.ts
```diff
@@ -0,0 +1,64 @@
...
+    credential_pop: credentialPop,
+  }).toString();
+
+  return await appFetch(statusAttUrl, {
+    method: "POST",
+    body,
+  })
+    .then(hasStatus(200))
+    .then((res) => res.json());
+};
```
grausof, line 64: This is an `application/json request`, not `application/x-www-form-urlencoded`

grausof (approved):
LGTM 🎉

PR #110, opened 2024-07-25
fix: Fix dPoP

grausof:
Fixes the use of DPoP and adds management of the dPoP crypto context. It should not introduce breaking change.
An ephemeral key is no longer used for each DPoP but the same key is recycled for each obtainment flow. Only at the end of obtaining the key is it deleted. If the flow is not successful, the same key can be reused.


LazyAfternoons (approved):
I don't like the idea of managing a key directly in this library as we use contexts and let the main app manage them. However given the time constraint we can merge this and review it later.

hevelius (approved):

PR #116, opened 2024-08-05
refactor: [SIW-1393] Use DPoP crypto context

LazyAfternoons:
#### List of Changes
- Change our flow functions to obtain a credential to accept a `dPopCryptoContext` which similarly to the `wiaCryptoContext` is used in the whole flow. This means the same key is used to generate DPoP tokens and it gets deleted only when flow ends. The key management is left to the consumer;
- Add `regenerateCryptoKey` which deletes the existing key if it exists and generates it again; 
- Add `deleteKeyIfExists` which deletes a crypto key if it exists via `io-react-native-wallet`, discarding the exception risen when it doesn't exist. This is useful when cleaning up keys after an error as they might not exists yet.

#### Motivation and Context
This PR is a follow up to #110 which was a hotfix but wasn't following the context pattern we've been using in this library.

Test for any regression in the existing flow, both with CIE+PIN and SPID/CieID.

#### Checklist:


example/src/scenarios/get-crendential.ts
```diff
@@ -146,5 +151,9 @@ export default (
     } catch (e) {
       console.error(e);
       return error(e);
+    } finally {
+      // Clean up ephemeral keys
+      deleteKeyIfExists(WIA_KEYTAG);
+      deleteKeyIfExists(DPOP_KEYTAG);
```
grausof, line 157: Why delete these keys? The Wallet Attestation key must be kept for the entire duration of the Wallet Attestation in case resubmission is necessary.

LazyAfternoons: We are regenerating a new wallet instance for each operation here. We used a random uuid before this change so that was basically lost anyway after the flow is done.

grausof: Ok I would add a comment to say that in the production implementation it should not be eliminated

grausof (approved):
LGTM

example/src/scenarios/get-crendential.ts
LazyAfternoons: Done in 2a50d9b5a4b22c37128ef4ad32f8c1d8362c34dd, thanks.

PR #117, opened 2024-08-06
chore: [SIW-1488] allow AbortSignal in completeUserAuthorizationWithQueryMode

gispada:
#### List of Changes

- `completeUserAuthorizationWithQueryMode` accepts an AbortSignal
- Created `OperationAbortedError`

```js
const abortController = new AbortController();

const { code } = await Credential.Issuance.completeUserAuthorizationWithQueryMode(
    issuerRequestUri,
    clientId,
    issuerConf,
    idphint,
    redirectUri,
    authorizationContext,
    abortController.signal
);

// Later...
abortController.abort();
```

#### Motivation and Context

In `completeUserAuthorizationWithQueryMode`, when opening the default browser with ReactNative's Linking module it is not currently possible to safely abandon the operation before the 120 seconds timeout. The listener on the "url" event is never cleared, resulting in a double invocation with stale data if the function is called again and the "url" event is received within those 120 seconds.

In order to fix this issue, the caller of the function can provide an [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) that races the promise to get the callback url. The first event that occurs resolves the race and all the listeners are removed.

I found it easier to test it directly on `io-app`, first without passing a signal to ensure no regressions were introduced, then with the signal and aborting/retrying the flow.

On the example app it should be possibile to successfully run the CieID scenario without regressions if no signal is provided.

#### Checklist:


LazyAfternoons:
I changed the base branch and remove the dependency from my PR as this definitely supersedes it. Thanks.

src/credential/issuance/04-complete-user-authorization.ts
```diff
@@ -100,13 +111,16 @@ export const completeUserAuthorizationWithQueryMode: CompleteUserAuthorizationWi
         });
     } else {
       // handler for redirectUri
-      Linking.addEventListener("url", ({ url }) => {
+      const sub = Linking.addEventListener("url", ({ url }) => {
```
grausof, line 114: what do you think about renaming `sub` to `eventListener` or something more understandable?

grausof (approved):

PR #118, opened 2024-08-06
feat: [SIW-1466] Add io backend login

grausof:
Adds login to io-backend via webview with idp choice. This way you can use production IO session without changing bearer token every time.
Thanks to @fdammassa @LazyAfternoons


LazyAfternoons (approved):
LGTM

PR #119, opened 2024-08-07
feat: [SIW-1496] Add navigation in the example app

LazyAfternoons:
> [!WARNING]
> Depends on #118

#### List of Changes
- Add necessary dependencies for `react-navigation`;
- Add a main native stack navigator with two routes. The home screen which is protected by the IO auth token and is conditionally rendered only when it is not `undefined`. The login screen which is rendered the the IO auth token is `undefined` in order to obtain one;
- Minor adjustments needed to render the screens; 
- [EXTRA] Replace `typecheck` with `tsc` which is mapped to `tsc --noEmit` in `package.json` to avoid running a `tsc` command and having multiple `.js` unwanted files;

#### Motivation and Context
This PR adds navigation to the example app. This is needed to have a more structured example app which can improve the developer experience and can also be distributed for testing.
Navigation allows having multiple screens instead of a single entry point which conditionally renders different screens.

Check for regression in the existing flows.

#### Checklist:


gispada (approved):
LGTM, the authenticated routes work as expected.

As a side note, it is not entirely clear to me why you renamed `typecheck` to `tsc`.

LazyAfternoons:
> LGTM, the authenticated routes work as expected.
> 
> As a side note, it is not entirely clear to me why you renamed `typecheck` to `tsc`.

Because running `yarn tsc` like I'm used to on `io-app` emits compiler output files like JavaScript source code, polluting the workspace. If that doesn't sound good to you we can revert it.

mastro993 (approved):

PR #120, opened 2024-08-10
feat: [SIW-1493] Add persistence and side effect management

LazyAfternoons:
> [!WARNING]
> Depends on #119 

> [!NOTE] 
> Given the size of this PR, some things have been descoped and will change in future PRs: 
> - Design refactor along with feedback on what's been persisted and what are the side effects of overriding the persisted state by rerunning a scenario;
> - Use navigation to manage different sections state; 
> - Abort operations; 
> - Better debug information on screen and in clipboard; 
> - CIE + PIN login refactor to align them to a standard scenario; 
> - Remove duplicate codes from scenarios (now thunks); 

#### List of Changes

- Refactor the whole app by using `Redux` in combination with `redux-thunk` to manage async operations. In order to reduce boilerplate and complexity I didn't add another middleware beside `redux-thunk` which is included by default in Redux Toolkit which is the suggested way to bootstrap a [Redux project](https://redux.js.org/introduction/getting-started). 
- Add slices for each macro scenario: attestation, credential, instance and issuance. Each slice has a portion of the store dedicated to the relative async operation required, thus carrying the `isLoading`, `isDone`, `hasError` information inside its state along with the result;
- Add thunks for async operations which reflect the slices structure; 
- Refactor `TestScenario` to encapsulate the logic behind an async operation, managing its state; 
- Refactor `TestCieL3Scenario` as above;
- Add persistence of the credentials (even though there's still no visuali feedback for it); 
- Refactor folder structure to accomodate the new store;
- Refactor the integrity context utils due to an issue with strict type checking.

#### Motivation and Context
This PR is needed to improve the developing experience of this library by managing side effects and persistence. 
This allows a better management of the flows within the example app, without the need to run each scenario every time the app is closed or Metro hot reloads.

Test for any regression in the existing scenarios. 
You can also test the persistence by 
- Register a wallet instance, close the app and try to obtain a wallet attestation or a PID directly;
- Obtain a PID, closing the app and then obtaining a credential directly.

#### Checklist:


grausof (approved):
Very good. I trust you

PR #128, opened 2024-08-29
[SIW-1508] Add TS scenario to the example app

mastro993:
#### List of Changes

This PR adds the ability to obtain the Health Insurance card credential

Open the example app, after obtaining a PID tap on the TS scenario and check that you are able to get the credential

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #129, opened 2024-08-30
feat: [SIW-1497] PRE-PROD env switch at runtime

LazyAfternoons:
> [!WARNING]
> Known issues: 
> - CIE + L3 flow is currently broken; 
> - Debug info slows the app when showing a DC/TS due to the long text;
> - The `Create Wallet Instance` scenario doesn't show a loading spinner.

> [!NOTE]
> In order to test the PRE environment you have to run through a proxy to inject the IPZS authentication header in the requests to IPSZ.

#### List of Changes
- Add both PRE-PROD environment; 
- Separate PID state from a generic credential for clarity due to the specific characteristics of the PID flow, including different auth methods. This is needed in order to support the IDP selection in PROD while obtaining a PID;
- Add support for the IDP selection when obtaining a PID with SPID;
- Add settings section;

#### Motivation and Context
This change allows hotswapping the environment while using the app, thus not requiring a new build when changing environments. 
This improves the developer experience but also allows us to distribute the example app internally. 

Static tests and tests of the existing flows.

#### Checklist:


grausof (approved):
Static check looks OK. Nothing to report. Really great work 🎉

mastro993:
> Debug info slows the app when showing a DC/TS due to the long text;

Can we fix this by truncating the content inside `DebugPrettyPrint`? We may still leave the entire content available for copying to the clipboard.

LazyAfternoons:
> > Debug info slows the app when showing a DC/TS due to the long text;
> 
> Can we fix this by truncating the content inside `DebugPrettyPrint`? We may still leave the entire content available for copying to the clipboard.

Yes, I'm already tracking the issue with a dedicated ticket as it's out of scope of this PR

PR #134, opened 2024-09-09
docs: [SIW-1529] Update documentation

LazyAfternoons:
#### List of Changes
Updates or creates the `README.md` of the main repository, the example app and each flow.

#### Motivation and Context
This PR adds a bit of documentation to the main repository, the example app and the library flows. This should make clear how flows should be structured, along with schemas and examples. The main goal is to give an overview given that the code is already well documented. 

You can check the preview by navigating the repo pointing at [this branch](https://github.com/pagopa/io-react-native-wallet/tree/SIW-1529-update-documentation).

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # 🪪 @pagopa/io-react-native-wallet
 
-Library which provides a high level abstraction to interact with the IT-Wallet ecosystem via a predefined flows, a set of utilities and helpers. It is designed to be used in [io-app](https://github.com/pagopa/io-app).
+Library which provides a high level abstraction to interact with the IT-Wallet ecosystem via a predefined flows, a set of utilities and helpers.
 Follows the [eudi-wallet-it-docs](https://github.com/italia/eudi-wallet-it-docs) specifications, currently aligned with version [0.7.1](https://github.com/italia/eudi-wallet-it-docs/releases/tag/0.7.1).
 
 ## Dependencies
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,251 +1,152 @@
 # 🪪 @pagopa/io-react-native-wallet
 
-📲 Provide data structures, helpers, and API to Wallet Instance.
+Library which provides a high level abstraction to interact with the IT-Wallet ecosystem via a predefined flows, a set of utilities and helpers. It is designed to be used in [io-app](https://github.com/pagopa/io-app).
+Follows the [eudi-wallet-it-docs](https://github.com/italia/eudi-wallet-it-docs) specifications, currently aligned with version [0.7.1](https://github.com/italia/eudi-wallet-it-docs/releases/tag/0.7.1).
```
hevelius, line 4: Perhaps it might be convenient to use an automatic badge that points to the latest release instead of having to update the readme manually. What do you think? https://github.com/badges/shields/issues/4293#issuecomment-549178270

example/README.md
```diff
@@ -0,0 +1,83 @@
+## Example
+
+This example app demonstrates how to use the `io-react-native-wallet` library to interact with the IO Wallet.
```
hevelius, line 3: IO Wallet is the internal project name. What do you think to use a more generic `Digital Identity Wallet`?

src/cie/README.md
```diff
@@ -0,0 +1,6 @@
+# CIE
+
+This library provides a components and a set of utilities to interact with the physical [CIE (Carta d'Identità Elettronica)](https://www.cartaidentita.interno.gov.it/) card. It can be used to [obtain a eID](../credential/issuance/README.md) via strong authentication.
```
hevelius, line 3:
```suggestion
This library provides a components and a set of utilities to interact with the physical [CIE (Carta d'Identità Elettronica)](https://www.cartaidentita.interno.gov.it/) card. It can be used to [obtain an eID](../credential/issuance/README.md) via strong authentication.
```

hevelius (changes requested):
Thank you very much @LazyAfternoons for the great job <3

README.md
LazyAfternoons: I think that might not be accurate since we are not always aligned to the latest release, what to you think?

example/README.md
LazyAfternoons: The example app does in fact interact with our internal project as it's supposed to log into IO and communicates with io-backend. My idea was to put emphasis on that.

hevelius (approved):

PR #136, opened 2024-09-10
feat: [SIW-1558] Add not entiled error when requesting a credential

LazyAfternoons:
#### List of Changes
- Adds a specific error `CredentialNotEntitledError` when the credential issuer returns `404` in response to the `/credential` request;
- Adds a generic error `CredentialRequestError` for generic errors.

#### Motivation and Context
This PR adds a specific error which is raised when `/credential` response contains a `404` code which indicates the user is not entitled to receive the requested credential.

With a proxy map locally the `/credential` response to a `404` and check in the debug menu of the example app if the custom error is thrown.

#### Checklist:


mastro993 (approved):
LGTM

PR #138, opened 2024-09-19
chore: [SIW-1635] Handle 201 status code when obtaining credential

gispada:
#### List of Changes

- Handled 201 status code in `obtainCredential`
- Created `CredentialIssuingNotSynchronousError`

#### Motivation and Context

This PR handle the 201 status code when invoking the `obtainCredential` function: this status code means the credential will be issued asynchronously, as per `IO-WALLET-DR-0021` design review.

Although the 201 status code is not an error, it has been handled as such in order to keep the return value of `obtainCredential` unchanged. Otherwise, the return value should differentiate between a requested credential and a full credential response.

A proxy was used to force a 201 status code.


LazyAfternoons (approved):
LGTM

PR #139, opened 2024-09-25
chore: [SIW-1669] Remove dev-proxy from the example app

LazyAfternoons:
#### List of Changes
- Remove the dev-proxy from `example` folder.

#### Motivation and Context
This PR removes the dev-proxy as it's not needed anymore. Everything on the example app goes through `io-backend` by logging in with SPID.

N/A

#### Checklist:


LazyAfternoons:
> It makes sense to me to delete it. The current setup only works with io-backend (prod) right?

That's correct. There's an issue with CodeQL I'm trying to fix though.

hevelius (approved):
LGTM. Fix static checks before merge.

PR #141, opened 2024-09-25
fix: [SIW-1679] Restore optional parameter to get all disclosures

gispada:
#### List of Changes

- Pass `includeUndefinedAttributes` to `parseCredentialSdJwt`.

#### Motivation and Context

Currently it is possible to pass `includeUndefinedAttributes` to `verifyAndParseCredential`, but that parameter is lost in the internal function invocations, so it is not possible to get disclosures/claims that are not explicitly supported in the entity configuration.

### DIFF
--- a/src/credential/issuance/07-verify-and-parse-credential.ts
+++ b/src/credential/issuance/07-verify-and-parse-credential.ts
@@ -215,6 +215,8 @@ const verifyAndParseCredentialSdJwt: WithFormat<"vc+sd-jwt"> = async (
  * @param credential The encoded credential returned by {@link obtainCredential}
  * @param format The format of the credentual returned by {@link obtainCredential}
  * @param context.credentialCryptoContext The crypto context used to obtain the credential in {@link obtainCredential}
+ * @param context.ignoreMissingAttributes Skip error when attributes declared in the issuer configuration are not found within disclosures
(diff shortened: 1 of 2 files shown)


LazyAfternoons (changes requested):
Should we update the function documentation with this new parameter and possibly also `ignoreMissingAttributes` which seems missing? I'd also add a note which specifies that they must be false in production.

gispada:
I have improved the documentation in cf2e9167800857caf3deb992a600e4a5c4557786.

LazyAfternoons:
> I have improved the documentation in [cf2e916](https://github.com/pagopa/io-react-native-wallet/commit/cf2e9167800857caf3deb992a600e4a5c4557786).

Good job, I'd also add it to the parameters of `verifyAndParseCredential`, something like: 
```
/**
 * Verify and parse an encoded credential.
...
 * @param context.ignoreMissingAttributes ...
 * @param context.includeUndefinedAttributes ...
...
 */
 ```
 
 Does it make sense to you?

gispada:
Updated `verifyAndParseCredential` JSDoc  in d2fc59c037f3acceef92daa736eff0f0c78253eb.

LazyAfternoons (approved):
LGTM

PR #142, opened 2024-09-26
fix: [SIW-1681] Change pattern to find the CIE authentication link

gispada:
#### List of Changes

- Changed the pattern to find the CIE authentication link (`a` element) to "lettura carta"
- Set the `a` text to lower case before matching the pattern

#### Motivation and Context

The CIE + PIN authentication method stopped working after the link text was changed in the CIE web app.

Tested on `io-app` to get a wallet instance.


LazyAfternoons (approved):
LGTM

PR #143, opened 2024-09-27
chore: [SIW-1684] Expose `iat` and `exp` claims

gispada:
#### List of Changes

- `verifyAndParseCredential` returns `iat` and `exp` claims as `issuedAt` and `expiration` respectively

#### Motivation and Context

We need to know the expiration and issuance date of credentials. The `exp` claim is included in the JWT payload, the `iat` is found among disclosures.

This PR attempts to expose those claims in a simple and backward-compatible way, adding them to the object returned by `verifyAndParseCredential`. Since this function is always called at the end of the issuance flow, it is convenient to also get some additional metadata without requiring another step of verification/decode later.

The use of `includeUndefinedAttributes` was discarded to avoid getting claims not declared in the issuer configuration.

#### Checklist:

### DIFF
--- a/src/credential/issuance/07-verify-and-parse-credential.ts
+++ b/src/credential/issuance/07-verify-and-parse-credential.ts
@@ -1,10 +1,11 @@
+import type { CryptoContext } from "@pagopa/io-react-native-jwt";
 import type { Out } from "../../utils/misc";
 import type { EvaluateIssuerTrust } from "./02-evaluate-issuer-trust";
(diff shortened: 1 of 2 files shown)


LazyAfternoons (commented):
Instead of returning it inside the flow, can't we expose a utility function which extract these claims? I would argue that it would be better to keep them separate, as they verifyAndParseCredential step should only verify and parse it. Extracting a specific claim should be done separately.

LazyAfternoons (approved):

PR #144, opened 2024-09-27
chore: [SIW-1632] Reuse the Wallet Instance Attestation in the example app

mastro993:
#### List of Changes

This PR removes the need to get a new Wallet Instance Attestation every time there is a credential issuance.
Once the WIA is obtained, it is stored and persisted to the device via Secure Storage and then used for the credential issuance.  The WIA is obtained again only when it expires.
This PR also introduces Secure Storage for WIA, PID and Credentials persistance.

#### Motivation and Context

Since new WIA requests on Android requires a Play Integrity API call, this change is required to reduce the Play Integrity API usage.

Run the example app and check that the WIA is requested only once or when it is expired

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #145, opened 2024-10-02
chore: [SIW-1707] Handle undefined `iat`claim

gispada:
#### List of Changes

- `verifyAndParseCredential` can return an undefined `issuedAt`

#### Motivation and Context

Although SD-JWTs should always contain the `iat` claim, we cannot have the absolute certainty. This PR takes a safe approach and handles a missing `iat` claim among disclosures.

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #147, opened 2024-10-03
feat: [SIW-1591] Add 409 handling with custom exception for wallet provider

LazyAfternoons:
#### List of Changes
- Update the `wallet-provider.yaml` specs due to changes in how the errors are mapped for certain endpoints like `/token` and `/wallet-instance`; 
- Create custom exceptions which are thrown when an error related to the integrity checks for `/token` and `/wallet-instances` fails, returning a `409` HTTP code; 
- Map these errors in their relative flows step for registering a wallet instance and obtaining an attestation.

#### Motivation and Context
This changes allows the consumer app to receive a custom error when an integrity check fails, thus allowing a specific text copy to be shown to the user.

- Open the example app and login; 
- Start a proxy like `Proxyman`;
- Try to register a wallet instance and obtain an attestation, everything should work just as before;
- Manually map the result of `/token` and `/wallet-instances` to `409 Conflict`;
- Try again to register a wallet instance and obtain an attestation. An error should occur. Check in the debug menu that the error is the one supposed to be thrown in this specific case.

#### Checklist:

### DIFF
--- a/openapi/wallet-provider.yaml
+++ b/openapi/wallet-provider.yaml
@@ -1,7 +1,11 @@
 openapi: 3.0.3
 info:
-  title: IO Wallet - User API
+  title: IO Wallet
   version: 1.0.0
+servers:
+  - url: https://api-app.io.pagopa.it/api/v1/wallet
+security:
+  - Bearer: []
 paths:
   /nonce:
     get:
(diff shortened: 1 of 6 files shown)


grausof (approved):
perfect

LazyAfternoons:
> Remember to increase the package version.

Usually I bump the versioni in a different commit, I prefer to do so

PR #149, opened 2024-10-16
feat: [SIW-1753] Add `authentic_source` field in supported credential metadata

gispada:
#### List of Changes

- Update zod schema to add  `authentic_source` in `SupportedCredentialMetadata`

#### Motivation and Context

It is necessary to display the authentic source in the credential's details. However this information is not found within disclosures, so this PR implements a solution that adds the authentic source in the entity configuration, for each supported credential.

#### Checklist:


LazyAfternoons (commented):
Why don't we put this as optional, just to be extra sure when parsing the EC? This also keep us covered if it will be implemented in the PROD env after we release the app.

gispada:
> Why don't we put this as optional, just to be extra sure when parsing the EC? This also keep us covered if it will be implemented in the PROD env after we release the app.

You are right, I wanted to avoid too many optional properties but it's better to be safe. Made the field optional in d98abdc5884833ffc029b1b001ec25fc5e5a1bea.

LazyAfternoons:
> > Why don't we put this as optional, just to be extra sure when parsing the EC? This also keep us covered if it will be implemented in the PROD env after we release the app.
> 
> You are right, I wanted to avoid too many optional properties but it's better to be safe. Made the field optional in [d98abdc](https://github.com/pagopa/io-react-native-wallet/commit/d98abdc5884833ffc029b1b001ec25fc5e5a1bea).

You can flag this for removal if you want, we already have a ticket for this: 
```// TODO [SIW-1268]: should not be optional```

gispada:
> You can flag this for removal if you want, we already have a ticket for this: `// TODO [SIW-1268]: should not be optional`

Let's keep it optional, it should be a temporary solution from my understanding.

LazyAfternoons (approved):
LGTM

PR #151, opened 2024-10-23
chore: [SIW-1779] Handle error codes during credential issuance and status attestation

gispada:
#### Motivation and Context

This PR creates the `CredentialInvalidStatusError` with `errorCode` to track specific errors in 403 and 404 responses during credential and status attestation requests, so that library consumers can extract more details on the failure. Ideally, this error code should be matched with one of the codes found in `issuance_errors_supported` in the EC to get the corresponding message.

> [!NOTE]
> The `CredentialNotEntitledError` and `StatusAttestationInvalid` have been superseded by `CredentialInvalidStatusError` as both those cases can be handled with a single error. 

#### List of Changes

- Renamed `StatusAttestationInvalid` to `CredentialInvalidStatusError` and added `errorCode`
- Removed `CredentialNotEntitledError`
- Added the response body in text format to `UnexpectedStatusCodeError`, otherwise the response body was lost (since this might be an unexpected error, JSON parsing the body might not be 100% safe)
- Throw `CredentialInvalidStatusError` in credential and status attestation requests
- Added an utility function to extract the error message from the EC

- Use a proxy to rewrite the response from the `/status` endpoint. It should return 404 with a body like `{ "error": "credential_revoked" }`.
- Do the same for the `/credential` endpoint. It should return 403 and 404 with a body like `{ "error": "credential_revoked" }`.

#### Checklist:

### DIFF
--- a/src/credential/issuance/06-obtain-credential.ts
+++ b/src/credential/issuance/06-obtain-credential.ts
@@ -5,16 +5,16 @@ import {
 } from "@pagopa/io-react-native-jwt";
 import type { AuthorizeAccess } from "./05-authorize-access";
 import type { EvaluateIssuerTrust } from "./02-evaluate-issuer-trust";
-import { hasStatus, type Out } from "../../utils/misc";
+import { hasStatus, safeJsonParse, type Out } from "../../utils/misc";
 import type { StartUserAuthorization } from "./03-start-user-authorization";
 import {
+  CredentialInvalidStatusError,
   CredentialIssuingNotSynchronousError,
-  CredentialNotEntitledError,
   CredentialRequestError,
   UnexpectedStatusCodeError,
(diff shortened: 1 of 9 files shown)


LazyAfternoons (commented):
Should we add a utility function here that extracts the error from the EC? What do you think?

gispada:
> Should we add a utility function here that extracts the error from the EC? What do you think?

It might come in handy. I added `extractErrorMessageFromIssuerConf` in c31e6006ff1fe36995a99694deccce0fae3fa022, that returns the error message with the shape of
```js
const message = {
  "it-IT": {
    title: "Error",
    description: "Description"
  }
}
```

What do you thinl?

grausof (approved):

mastro993 (approved):

PR #152, opened 2024-11-05
fix: [SIW-1795] Problem with SPID Login Using Custom Tab

RiccardoMolinari95:
#### List of Changes

- Renamed the screen `PidSpidLoginScreen` to `PidSpidIdpSelectionScreen`. When an IDP is selected, the authorization flow starts, and the user is redirected to the new `PidSpidLoginScreen` after obtaining the `authUrl`.

- Created the new screen `PidSpidLoginScreen`, which uses a `WebView` to load `authUrl`. Once the authorization is completed and the redirect URL is obtained, the flow continues to complete the PID issuance.

- Updated the `pid.ts` file to manage authentication via SPID and CIE L3, including the creation of two new thunks: `preparePidFlowParamsThunk` and `continuePidFlowThunk`. `preparePidFlowParamsThunk` obtains the `authUrl`, passes it to the respective `WebView` for rendering, and saves the necessary parameters for the next step in the state. `continuePidFlowThunk` receives the redirect URL from the WebView and continues the process to obtain the PID.

- The parameters PreparePidFlowParamsThunkInput and PreparePidFlowParamsThunkOutput have been set to handle both the CIE and SPID flows, so some parameters can be optional, such as ciePin.

-  Added `openUrlAndListenForAuthRedirect`, which opens the `authUrl` in the browser for CIE ID and provides the `authRedirectURL`. This step was previously handled by `CompleteUserAuthorizationWithQueryMode`

- Added the function `buildAuthorizationUrl` in step `04-complete-user-authorization`, which constructs the `authUrl` that will be passed to the WebView for CIE L3 and SPID, and to `openUrlAndListenForAuthRedirect` for CIE ID

- The parameters `issuerRequestUri`, `clientId`, `issuerConf`, and `idpHint` have been removed from the type `CompleteUserAuthorizationWithQueryMode`. Now `CompleteUserAuthorizationWithQueryMode` receives the `authRedirectURL` from the WebView for CIE L3 or SPID, or from the browser redirect for CIE L2, and performs the parsing.

#### Motivation and Context

This PR removes the usage of the `io-react-native-login-utils` library and its `Custom Tab `implementation for SPID authentication to obtain the PID. The `Custom Tab` has been replaced with a `WebView` to avoid issues that some users were experiencing during the SPID login process.
The login flow via SPID has been unified with the CIE L3 flow due to their similarities and has been split into two parts. The CIE L2 flow remains the same as before.

Successfully obtained the PID using all three authentication methods: SPID, CIE L2, and CIE L3.

https://github.com/user-attachments/assets/ba5af9fb-20da-4184-83c4-8cd0e85b50eb

#### Checklist:


LazyAfternoons (changes requested):
Good work, I'd like to point out a few things we might refine, if that makes sense to you: 
- `startUserAuthroization` should keep the same implementation as before to keep it consistent between eID and a generic credential. This allow us to have a strict signature in terms of input/output without the need of declaring the `authUrl` which might not be needed in case of a generic credential; 
- We should create a new function in the `04-complete-user-authroization` step which builds the authentication urls and will be only used in case of a eID issuance, similarly 
to `getCredentialToBePresented` which is used only for other credentials; 
- In the `completeUserAuthorizationWithQueryMode` we should only parse the result redirect url which comes from the consumer who performed the strong authentication; 
- We should move the CieID logic to the consumer app since we don't manage any kind of authentication directly in the library anymore.

Let me know what you think. 

```mermaid
graph TD;
    2[...]
    3[startUserAuthorization]
    C4[getRequestedCredentialToBePresented]
    C4.1[completeUserAuthorizationWithFormPostJwtMode]
    E4[getAuthorizationUrl]
    E4.1[Consumer performs strong authentication via SPID/CIE with authroization URL]
    E4.2[completeUserAuthorizationWithQueryMode]
    5[authorizeAccess]
    6[..]
    credSel{Is credential an eID?}
    
    2 --> 3
    3 --> credSel
    credSel -->|Yes| E4
    credSel -->|No| C4
    C4 --> C4.1
    C4.1 --> 5
    E4 --> E4.1
    E4.1 --> E4.2
    E4.2 --> 5
    5 --> 6
```

RiccardoMolinari95:
> Good work, I'd like to point out a few things we might refine, if that makes sense to you:
> 
> * `startUserAuthroization` should keep the same implementation as before to keep it consistent between eID and a generic credential. This allow us to have a strict signature in terms of input/output without the need of declaring the `authUrl` which might not be needed in case of a generic credential;
> * We should create a new function in the `04-complete-user-authroization` step which builds the authentication urls and will be only used in case of a eID issuance, similarly
>   to `getCredentialToBePresented` which is used only for other credentials;
> * In `completeUserAuthorizationWithQueryMode` we should only parse the result redirect url which comes from the consumer who performed the strong authentication;
> * We should move the CieID logic to the consumer app since we don't manage any kind of authentication directly in the library anymore.
> 
> Let me know what you think.
> 
> ```mermaid
> graph TD;
>     2[...]
>     3[startUserAuthorization]
>     C4[getRequestedCredentialToBePresented]
>     C4.1[completeUserAuthorizationWithFormPostJwtMode]
>     E4[getAuthorizationUrl]
>     E4.1[Consumer performs strong authentication via SPID/CIE with authroization URL]
>     E4.2[completeUserAuthorizationWithQueryMode]
>     5[authorizeAccess]
>     6[..]
>     credSel{Is credential an eID?}
>     
>     2 --> 3
>     3 --> credSel
>     credSel -->|Yes| E4
>     credSel -->|No| C4
>     C4 --> C4.1
>     C4.1 --> 5
>     E4 --> E4.1
>     E4.1 --> E4.2
>     E4.2 --> 5
>     5 --> 6
> ```

Everything makes sense to me! I also prefer not to use optional parameters.
I will proceed with making the changes.

src/index.ts
```diff
@@ -12,6 +12,7 @@ import * as WalletInstanceAttestation from "./wallet-instance-attestation";
 import * as Trust from "./trust";
 import * as WalletInstance from "./wallet-instance";
 import * as Cie from "./cie";
+import * as Misc from "./utils/misc";
```
LazyAfternoons, line 15: Do we need this export?

RiccardoMolinari95: Yes, it makes sense

src/credential/issuance/__tests__/04-complete-user-authorization.test.ts
```diff
@@ -1,48 +1,23 @@
...
-    const result = await completeUserAuthorizationWithQueryMode(
-      issuerRequestUri,
-      clientId,
-      issuerConf,
-      idpHint,
-      redirectUri,
-      authContext
-    );
+    const query = parseUrl(authRedirectUrl).query;
+    const authResParsed = parseAuthroizationResponse(query);
```
gispada, line 18: Is there a reason we are not testing `completeUserAuthorizationWithQueryMode` but only `parseAuthroizationResponse`?

src/credential/issuance/__tests__/04-complete-user-authorization.test.ts
```diff
@@ -51,44 +26,30 @@ describe("authorizeUserWithQueryMode", () => {
...
-        issuerConf,
-        idpHint,
-        redirectUri,
-        authContext
-      )
-    ).rejects.toThrowError(AuthorizationIdpError);
+    const query = parseUrl(authRedirectUrl).query;
+
+    try {
+      await parseAuthroizationResponse(query);
```
gispada, line 34: Same as above.

src/credential/issuance/__tests__/04-complete-user-authorization.test.ts
```diff
@@ -51,44 +26,30 @@ describe("authorizeUserWithQueryMode", () => {
...
-        issuerRequestUri,
-        clientId,
-        issuerConf,
-        idpHint,
-        redirectUri,
-        authContext
-      )
-    ).rejects.toThrowError(AuthorizationError);
+    try {
+      await parseAuthroizationResponse(query);
```
gispada, line 50: Same as above.

example/ios/IoReactNativeWalletExample.xcodeproj/project.pbxproj
gispada, line 1: Can you check why this file was modified? Usually this file should not change.

example/src/thunks/pidCieID.ts
gispada, line 1: Do we really need to have a separate flow for CieID? Would it be possible to handle all three scenarios the same way?
1. `preparePidFlowParamsThunk`
2. Open webview for Spid and CIE L3 / open browser for CieID -> get the auth redirect url
4. `continuePidFlowThunk`

src/credential/issuance/__tests__/04-complete-user-authorization.test.ts
RiccardoMolinari95: It is an oversight, thank you for noticing. I will correct it

src/credential/issuance/04-complete-user-authorization.ts
```diff
@@ -63,98 +48,55 @@ export type GetRequestedCredentialToBePresented = (
...
+  authUrl: string;
+}>;
+
 /**
- * WARNING: This function must be called after {@link startUserAuthorization}. The next function to be called is {@link authorizeAccess}.
- * The interface of the phase to complete User authorization via strong identification when the response mode is "query" and the request credential is a PersonIdentificationData.
- * It is used to complete the user authorization by catching the redirectSchema from the authorization server which then contains the authorization response.
- * This function utilizes the authorization context to open an in-app browser capable of catching the redirectSchema to perform a get request to the authorization endpoint.
- * If the 302 redirect happens and the redirectSchema is caught, the function will return the authorization response after parsing it from the query string.
+ * WARNING: This function must be called after {@link startUserAuthorization}. The generated authURL will be used in the webviews for CIE L3 and SPID, and in the function {@link openUrlAndListenForAuthRedirect} for CIEID.
```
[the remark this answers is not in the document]
RiccardoMolinari95: It is an oversight, thank you for noticing. I will correct it

example/src/thunks/pidCieID.ts
RiccardoMolinari95: Yes could be an option, for future improvement

src/credential/issuance/04-complete-user-authorization.ts
```diff
@@ -63,98 +48,55 @@ export type GetRequestedCredentialToBePresented = (
...
+  authUrl: string;
+}>;
+
 /**
- * WARNING: This function must be called after {@link startUserAuthorization}. The next function to be called is {@link authorizeAccess}.
- * The interface of the phase to complete User authorization via strong identification when the response mode is "query" and the request credential is a PersonIdentificationData.
- * It is used to complete the user authorization by catching the redirectSchema from the authorization server which then contains the authorization response.
- * This function utilizes the authorization context to open an in-app browser capable of catching the redirectSchema to perform a get request to the authorization endpoint.
- * If the 302 redirect happens and the redirectSchema is caught, the function will return the authorization response after parsing it from the query string.
+ * WARNING: This function must be called after {@link startUserAuthorization}. The generated authURL will be used in the webviews for CIE L3 and SPID.
```
gispada, line 61:
```suggestion
 * WARNING: This function must be called after {@link startUserAuthorization}. The generated authUrl must be used to open a browser or webview capable of catching the redirectSchema to perform a get request to the authorization endpoint.
```

example/src/thunks/pidCieID.ts
gispada: Can you create a Jira Task and add a TODO comment to reference that task? We can treat it as a low priority task, but I would like to remove code duplication, even if it's just an example app.

RiccardoMolinari95: Good idea, so we can keep track of the activity

example/src/store/reducers/pid.ts
```diff
@@ -103,75 +104,94 @@ const pidSlice = createSlice({
...
      * for the requested credential.
      */
-    builder.addCase(continueCieL3FlowThunk.pending, (state) => {
-      // Redundant as already set by prepareCieL3FlowParams but we want to be explicit and set the loading state
-      state.pidAsyncStatus.cieL3 = {
+    builder.addCase(continuePidFlowThunk.pending, (state) => {
+      // Redundant as already set by preparePidFlowParams but we want to be explicit and set the loading state
+      const cieL3IsLoading = state.pidAsyncStatus.cieL3.isLoading;
+      const spidIsLoading = state.pidAsyncStatus.spid.isLoading;
+      const authMethod = cieL3IsLoading
```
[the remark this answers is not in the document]
RiccardoMolinari95: I set "spid" as the default value in case neither `cieL3IsLoading` nor `spidIsLoading` is true. This ensures that `authMethod` always has a valid value, even if neither of the authentication methods is currently loading.

gispada (approved):
LGTM. I'd wait for another approval though, as this PR involves a high risk flow.

grausof (approved):
LGTM 🎉

mastro993:
On iOS, I'm unable to complete the PID issuance with SPID. After a successful login, I get stuck on this screen:
 


Am I missing something?

RiccardoMolinari95:
> On iOS, I'm unable to complete the PID issuance with SPID. After a successful login, I get stuck on this screen:
> 
> 
> Am I missing something?

Is this a mistake that you always have ? Have you tried deleting the example app from the phone and restarting from a clean situation ?

example/src/screens/login/PidLoginScreen.tsx
```diff
@@ -0,0 +1,106 @@
...
+    if (idpIntent) {
+      Linking.openURL(idpIntent);
+      return false;
+    }
+    return true;
+  };
+
+  const handleNavigationStateChange = async (navState: WebViewNavigation) => {
+    const { url } = navState;
+    if (url.includes("iowallet")) {
```
mastro993, line 54: `includes` it's too broad, we could use `startsWith` instead. What do you think?

```suggestion
    if (url.startsWith("iowallet")) {
```

RiccardoMolinari95: Thanks for the suggestions. i will implement them

example/src/thunks/pid.ts
```diff
@@ -1,66 +1,245 @@
...
+    ReturnType<typeof Credential.Issuance.startUserAuthorization>
+  >["codeVerifier"];
+  walletInstanceAttestation: Awaited<
+    ReturnType<typeof WalletInstanceAttestation.getAttestation>
+  >;
+  credentialDefinition: Awaited<
+    ReturnType<typeof Credential.Issuance.startUserAuthorization>
+  >["credentialDefinition"];
+  redirect_uri: string;
+  ciePin?: string;
```
grausof, line 68: Always use camel case

example/src/thunks/pid.ts
```diff
@@ -1,66 +1,245 @@
...
+
+  const isCie = args.idpHint.includes("servizicie") ? true : false;
+
+  const wiaCryptoContext = createCryptoContextFor(WIA_KEYTAG);
+
+  // Get env
+  const env = selectEnv(getState());
+  const { WALLET_PID_PROVIDER_BASE_URL, REDIRECT_URI } = getEnv(env);
+
+  const redirect_uri = isCie ? CIE_L3_REDIRECT_URI : REDIRECT_URI;
```
grausof, line 109:
```suggestion
  const redirectUri = isCie ? CIE_L3_REDIRECT_URI : REDIRECT_URI;
```

example/src/screens/login/PidLoginScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (idpIntent) {
+      Linking.openURL(idpIntent);
+      return false;
+    }
+    return true;
+  };
+
+  const handleNavigationStateChange = async (navState: WebViewNavigation) => {
+    const { url } = navState;
+    if (url.startsWith(REDIRECT_URI)) {
```
grausof, line 58: Avoid reading from env. You can get the redirectUri from `route.params`

example/src/screens/login/PidSpidIdpSelectionScreen.tsx
```diff
@@ -6,33 +6,41 @@ import IdpsGrid from "../../components/IdpsGrid";
...
-      );
+      try {
+        const { authUrl } = await dispatch(
+          preparePidFlowParamsThunk({
+            idpHint,
+            authMethod: "spid",
+            credentialType: "PersonIdentificationData",
+          })
+        ).unwrap();
+        navigation.navigate("PidSpidLogin", { authUrl });
```
grausof, line 40: here you can also pass redirectUri since you need it in WebView

grausof (changes requested):

example/src/screens/login/PidSpidIdpSelectionScreen.tsx
RiccardoMolinari95: Thanks, nice suggestions

example/src/screens/login/PidLoginScreen.tsx
```diff
@@ -0,0 +1,106 @@
...
+        androidCameraAccessDisabled={true}
+        androidMicrophoneAccessDisabled={true}
+        allowsInlineMediaPlayback={true}
+        mediaPlaybackRequiresUserAction={true}
+        originWhitelist={originSchemasWhiteList}
+        cacheEnabled={false}
+        onShouldStartLoadWithRequest={handleShouldStartLoading}
+        onNavigationStateChange={handleNavigationStateChange}
+        userAgent="Mozilla/5.0 (iPhone; CPU iPhone OS 14_0_1 like Mac OS X; Linux; Android 10) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.93 Mobile Safari/537.36"
+      />
```
grausof, line 84: What do you think about integrating this component into an ad-hoc io-react-native-wallet component like done for CieL3 WebView?

grausof (approved):
Great 🎉

mastro993 (approved):
LGTM

grausof:
This PR is ok for now. Need a separate PR to fix the sample app on iOS!

PR #155, opened 2024-11-08
chore: [SIW-1762] Refactor errors

gispada:
#### List of Changes

- Errors related to the interactions between the library and the issuer/wallet provider's backend have been split into one of two classes: `IssuerResponseError` and `WalletProviderResponseError`. Within these classes, the specific error can be found in the `code` property.
- The inheritance of error classes has been modified:
error-classes
- Removed differences between HTTP errors to improve consistency: they all have the same properties, with `reason` containing a specific reason when available or the response body.
- Moved specific errors to the feature folders they belong to.
- Throw `IssuerResponseError` in those phases of the issuance flow that are not explicitly handled.

>[!WARNING]
> This PR introduces breaking changes in the way errors are handled. The majority of errors have been deleted and superseded by `IssuerResponseError` and `WalletProviderResponseError`.

#### Motivation and Context

The main motivation for this refactor was to accurately track errors returned by the issuer and by the wallet provider, and let more information surface to the caller. Additionally HTTP errors were not always consistent, having different or unclear properties.

The HTTP errors were simulated using a proxy, and the logs inspected to ensure JS errors were the ones expected.

#### Checklist:

Co-authored-by: [Alessandro Mazzon](https://github.com/ale-mazz)


src/sd-jwt/errors.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * When claims are requested but not found in the credential
+ *
+ */
+export class ClaimsNotFoundBetweenDislosures extends Error {
```
mastro993, line 5:
```suggestion
export class ClaimsNotFoundBetweenDisclosures extends Error {
```

gispada: Fixed in 3efcbac70e93f89483c6cecdb0afbb3ffd3d93ff, along with other typos.

src/credential/issuance/05-authorize-access.ts
```diff
@@ -103,12 +103,15 @@ export const authorizeAccess: AuthorizeAccess = async (
...
+    .then(hasStatusOrThrow(200, IssuerResponseError))
     .then((res) => res.json())
     .then((body) => TokenResponse.safeParse(body));
 
   if (!tokenRes.success) {
-    throw new ValidationFailed(tokenRes.error.message);
+    throw new ValidationFailed({
+      message: "Token Response validation failed",
+      reason: tokenRes.error.message,
+    });
```
grausof, line 114: I see it duplicated many times. What do you think about writing a method for this?

src/credential/issuance/errors.ts
```diff
@@ -0,0 +1,47 @@
+import { IoWalletError, serializeAttrs } from "../../utils/errors";
+
+//TODO: Check for io-react-native-utils before removing them
```
grausof, line 3: What do you think about adding the linked jira ticket to the TODO?

src/credential/issuance/errors.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export class AuthorizationError extends IoWalletError {
+  code = "ERR_IO_WALLET_AUTHORIZATION_ERROR";
+
+  constructor(message?: string) {
+    super(message);
+  }
+}
+
+//TODO: Check for io-react-native-utils before removing them
```
grausof, line 15: ditto

src/credential/presentation/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(
+    message: string,
+    claim: string = "unspecified",
+    reason: string = "unspecified"
+  ) {
+    super(serializeAttrs({ message, claim, reason }));
+    this.claim = claim;
+    this.reason = reason;
+  }
+}
```
grausof, line 25: what do you think about?
```suggestion
export class AuthRequestDecodeError extends IoWalletError {
  static readonly CODE = "ERR_IO_WALLET_AUTHENTICATION_REQUEST_DECODE_FAILED";

  readonly claim: string;
  readonly reason: string;

  constructor(message: string, claim = "unspecified", reason = "unspecified") {
    super(serializeAttrs({ message, claim, reason }));
    this.claim = claim;
    this.reason = reason;
  }

  get code() {
    return AuthRequestDecodeError.CODE;
  }
}
```

src/sd-jwt/errors.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** The Claims not found */
+  claims: string[];
+
+  constructor(claims: string | string[]) {
+    const c = Array.isArray(claims) ? claims : [claims];
+    const message = `Some requested claims are not present in the disclosurable values, claims: ${c.join(
+      ", "
+    )}`;
+    super(message);
+    this.claims = c;
```
grausof, line 17:
```suggestion
      const claimsArray = Array.isArray(claims) ? claims : [claims];
      super(`Some requested claims are not present in the disclosurable values: ${claimsArray.join(", ")}`);
      this.claims = claimsArray;
```

src/utils/errors.ts
```diff
@@ -595,3 +198,74 @@ export function extractErrorMessageFromIssuerConf(
...
+ *   .handle("*", { code: "ERROR_CODE_3", message: "Fallback" })
+ *   .buildFrom(baseError)
+ * ```
+ */
+export class ResponseErrorBuilder<T extends typeof UnexpectedStatusCodeError> {
+  private errorCases: {
+    [K in number | "*"]?: {
+      code: ErrorCodeMap<T>;
+      message: string;
+      reason?: GenericErrorReason;
```
grausof, line 250: ditto. Maybe it's better to define a type?
```ts
type ErrorCase<T> = {
  code: ErrorCodeMap<T>;
  message: string;
  reason?: GenericErrorReason;
};
```

src/credential/presentation/errors.ts
gispada: This error was moved from another folder; the only modification was the removal of the static property. It's also in the presentation section, which is outdated, not used and will probably require future changes. If that's ok for you, I would leave the file as is.

src/credential/issuance/errors.ts
gispada: I removed the TODO: in the context of this PR these errors are perfectly fine. Whether to remove them or not depends on another PR and on the merge order of the two PRs. We'll see which one gets merged first.

src/sd-jwt/errors.ts
gispada: Addressed in 3efbcee36cac790849c39a455cac6c45ea8ce5e6.

src/client/index.ts
```diff
@@ -44,13 +44,7 @@ export const getWalletProviderClient = (context: {
...
       })
         .then(validateResponse)
-        .then((res) => {
-          const contentType = res.headers.get("content-type");
-          if (contentType?.includes("application/json")) {
-            return res.json();
-          }
-          return res.text();
-        }),
+        .then<string | Record<string, unknown>>(parseRawHttpResponse),
```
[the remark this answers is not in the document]
gispada: This explicit type here was need to solve a TS error.

About Zod validation, we don't know the exact shape of the response here; I think validation at this level should be handled by `typed-openapi`, because it's the one that parses the OpenAPI file and knows the shape of every response.

Anyway, we usually validate the raw response in the call site with Zod.

mastro993 (approved):
LGTM!

grausof (approved):

PR #156, opened 2024-11-16
feat: [SIW-1844] Add trustmark generation functions

mastro993:
#### List of Changes

- Added `getCredentialTrustmarkJwt` function, which returns the credential's trustmark JWT 
- Updated the example app with the screens, store and functions to test the new trustmark functionality

#### Motivation and Context

This approach addresses vulnerabilities in the previous static QR code implementation.
This new TrustMark implementation binds the QR code to the specific credential being verified, preventing QR code swapping and ensuring unique, tamper-proof authentication. It uses a signed JWT with a short validity period for enhanced security.

The trustmark JWT can be obtained using the following function, exported from `Credential.Trustmark`:

```typescript
function getCredentialTrustmark({
  walletInstanceAttestation: string,
  wiaCryptoContext: CryptoContext,
  credentialType: string,
  docNumber?: string,
  expirationTime?: number | string
}): Promise<{
  jwt: string,
  expirationTime: number
}>
```

Tested the Trustmark generation using the example app

https://github.com/user-attachments/assets/1ce7de8f-39a2-405a-88ce-50e29e3c6f76

#### Checklist:


example/src/screens/TrustmarkScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+ * Returns the document number for a credential, if applicable
+ * @param credential the credential from which to extract the document number
+ * @returns a string representing the document number, undefined if not found
+ */
+const getCredentialDocumentNumber = (
+  credential: CredentialResult | undefined
+): string | undefined => {
+  if (credential === undefined) {
+    return undefined;
+  }
```
grausof, line 145: Why not?
```suggestion
const getCredentialDocumentNumber = (
  credential: CredentialResult
): string | undefined => {
```

example/src/screens/TrustmarkScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  return credential.parsedCredential.document_number?.value as string;
+};
+
+/**
+ * Displsys a countdown timer for the QR Code expiration time
```
grausof, line 151:
```suggestion
 * Display a countdown timer for the QR Code expiration time
```

src/credential/trustmark/get-credential-trustmark.ts
```diff
@@ -0,0 +1,39 @@
...
+      iss: walletInstanceAttestation,
+      sub: credentialType,
+      subtyp: documentNumber,
+    })
+    .setIssuedAt()
+    .setExpirationTime("2m")
+    .sign();
+
+  return signedTrustmarkJwt;
+};
```
grausof, line 39: I suggest adding a runtime check to verify that the public key present in the walletInstanceAttestation is the one associated with wiaCryptoContext.

grausof (changes requested):

example/src/screens/TrustmarkScreen.tsx
mastro993: Addressed here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/558242937799b2b389e788c18168467c19b20dc4

src/credential/trustmark/get-credential-trustmark.ts
mastro993: Addressed here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/ae8c8c15c4b8ff1cc75c85fdcc377efb2b1f8c72

src/credential/trustmark/get-credential-trustmark.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  const holderBindingKey = await wiaCryptoContext.getPublicKey();
+  const decodedWia = WalletInstanceAttestation.decode(
+    walletInstanceAttestation
+  );
+
+  if (
+    !decodedWia.payload.cnf.jwk.kid ||
+    decodedWia.payload.cnf.jwk.kid !== holderBindingKey.kid
+  ) {
```
grausof, line 39: not for this PR, but usually a check on the kid alone might not be enough. The thumbprint of both keys should be calculated and compared.

grausof (approved):
LGTM

grausof (changes requested):

src/credential/trustmark/get-credential-trustmark.ts
```diff
@@ -0,0 +1,69 @@
+import { SignJWT, type CryptoContext } from "@pagopa/io-react-native-jwt";
+import { WalletInstanceAttestation } from "@pagopa/io-react-native-wallet";
```
grausof, line 2:
```suggestion
import { WalletInstanceAttestation } from "../../wallet-instance-attestation";
```

src/credential/trustmark/get-credential-trustmark.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  const signedTrustmarkJwt = await new SignJWT(wiaCryptoContext)
+    .setProtectedHeader({
+      alg: "ES256",
+    })
+    .setPayload({
+      iss: walletInstanceAttestation,
+      sub: credentialType,
+      subtyp: obfuscatedDocumentNumber,
+    })
```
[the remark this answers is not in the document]
mastro993: Addressed here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/487e7707ded53770221c2377974e36eb09547593

src/credential/trustmark/get-credential-trustmark.ts
mastro993: Addressed here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/cc872b1bcddc6900fd7b725ec558418dbb5365a9

src/credential/trustmark/get-credential-trustmark.ts
mastro993: I've updated the check here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/88bd9ee07565b9bdd43b0b2a219ee74f0eb397de
Let me know!

src/credential/trustmark/__tests__/get-credential-trustmark.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  it("should generate a JWT with the correct payload", async () => {
+    const ephemeralKeytag = `ephemeral-${Math.random()}`;
+    await generate(ephemeralKeytag);
+
+    const walletInstanceAttestation = "walletInstanceAttestation";
+    const wiaCryptoContext = createCryptoContextFor(ephemeralKeytag);
+    const credentialType = "MDL";
+    const documentNumber = "1234567890";
+
+    const jwt = await getCredentialTrustmarkJwt(
```
grausof, line 16: You have a WIA decode in `getCredentialTrustmarkJwt`. I don't think it can use "walletInstanceAttestation" as a mocked value

src/credential/trustmark/get-credential-trustmark.ts
grausof: Top 🎆

src/credential/trustmark/__tests__/get-credential-trustmark.test.ts
mastro993: Here we go! https://github.com/pagopa/io-react-native-wallet/pull/156/commits/3e46f0cd86bb316d0345ee65b00dffac21695ccd

grausof (approved):
Great 🔥

grausof:
Remember to increase the minor version in package.json before release

src/credential/trustmark/get-credential-trustmark.ts
```diff
@@ -0,0 +1,74 @@
...
+    .setPayload({
+      iss: walletInstanceAttestation,
+      sub: credentialType,
+      /**
+       * If present, the document number is obfuscated before adding it to the payload
+       */
+      ...(docNumber ? { subtyp: obfuscateString(docNumber) } : {}),
+    })
+    .setIssuedAt()
+    .setExpirationTime("2m")
```
grausof, line 70: almost done 😅, what do you think about making this function return the duration in seconds? This way whoever uses it knows how to implement the countdown without having hardcoded values. Or you could also think about passing the duration to the function that generates the trustmark

grausof (changes requested):

src/credential/trustmark/get-credential-trustmark.ts
mastro993: We can do both! Passing the duration as a parameter for the function (with a default value) could be a great addition. Additionally, returning the exact expiration time could save extra work for those using the function.

grausof: yes perhaps better to return the expiration timestamp by accessing the exp field of the JWT, even if it is still included inside but it would require parsing by the user

mastro993: Addressed here: https://github.com/pagopa/io-react-native-wallet/pull/156/commits/b6c4f17650e3e2102557a1faffceced0aa33e87e

If also updated the docs, tests and the example app to reflect the change. Let me know!

grausof (approved):
Now it's perfect!

PR #166, opened 2024-12-06
refactor: [EUDIW-202] Move CIE components to example app

LazyAfternoons:
#### List of Changes
- Move the whole CIE folder to the example app; 
- Remove `react-native-cie` and `react-native-webview` dependencies from the library.

#### Motivation and Context
Currently, the `io-react-native-wallet` package directly includes components related to the CIE . This forces the consumer app to install the associated dependencies even in cases where, like in the EUDIW application, login via CIE is not implemented at all. Since it is the responsibility of the app using the library to choose the authentication method for the user, this dependency needs to be removed.

Test a eID issuance flow with CIE l3.

#### Checklist:


mastro993 (approved):
LGTM, but we should merge this after https://github.com/pagopa/io-react-native-wallet/pull/165

PR #170, opened 2025-01-02
chore: [SIW-1310] Generation of cryptographically secure UUIDs

RiccardoMolinari95:
This PR replaces the library used for UUID generation. The `react-native-uuid` library, which relies on `Math.random`, is replaced with the `uuid` library, which utilizes `Crypto.getRandomValues()` for cryptographically secure random values.

#### List of Changes

- Added `react-native-get-random-values` polyfill to use `Crypto.getRandomValues()`

#### Motivation and Context

This change is required to enhance the security of UUID generation. The previous library, `react-native-uuid`, uses a pseudo-random generator based on `Math.random`, which is not suitable for cryptographic purposes. The new library, `uuid`, uses `Crypto.getRandomValues()`, providing a higher level of randomness and security.

All functionality should work as before

#### Checklist:

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -1,3 +1,4 @@
+import 'react-native-get-random-values';
 import { AppRegistry } from "react-native";
 import App from "./src/App";
 import { name as appName } from "./app.json";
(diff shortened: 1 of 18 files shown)


mastro993 (approved):
LGTM, tested on iOS and everything work as expected. Great job

PR #174, opened 2025-01-15
feat: [EUDIW-209] Add Remote presentation for Potential specs

manuraf:
#### Checklist:


src/credential/presentation/05-verify-request-object.ts
```diff
@@ -0,0 +1,38 @@
...
+      pubKey = jwkKeys.find(
+        ({ kid }) => kid === requestObjectJwt.protectedHeader.kid
+      );
+    }
+
+    if (!pubKey) {
+      throw new UnverifiedEntityError("Request Object signature verification!");
+    }
+    await verify(requestObjectEncodedJwt, pubKey);
+
```
LazyAfternoons, line 29: Usually we avoid the usage of `let` as much as possibile to prefer an immutable data approach. This could be refactored as: 
```
const pubKey = jwkKeys?.find(
  ({ kid }) => kid === requestObjectJwt.protectedHeader.kid
);

if (!pubKey) {
  throw new UnverifiedEntityError("Request Object signature verification!");
}
```

What do you think?

src/credential/presentation/07-evaluate-input-descriptor.ts
```diff
@@ -0,0 +1,235 @@
...
+  (inputDescriptor, payloadCredential, disclosures) => {
+    if (!inputDescriptor?.constraints?.fields) {
+      // No validation, all field are optional
+      return {
+        requiredDisclosures: [],
+        optionalDisclosures: disclosures,
+      };
+    }
+    let requiredClaimNames: string[] = [];
+    let optionalClaimNames: string[] = [];
```
LazyAfternoons, line 158: Same as above

src/credential/presentation/08-send-authorization-response.ts
```diff
@@ -0,0 +1,252 @@
...
+      vp_token,
+      presentation_submission
+    );
+  } else {
+    requestBody = await buildDirectPostBody(
+      requestObject,
+      vp_token,
+      presentation_submission
+    );
+  }
```
LazyAfternoons, line 239: Same as above

src/credential/presentation/07-evaluate-input-descriptor.ts
```diff
@@ -0,0 +1,235 @@
...
+    // For each field, we need at least one matching path
+    // If we succeed, we push the matched disclosure in matchedDisclosures and stop checking further paths
+    const allFieldsValid = inputDescriptor.constraints.fields.every((field) => {
+      // For Potential profile, selectively disclosed claims will always be built as an individual object property, by using a name-value pair.
+      // Hence that selective claim for array element and recursive disclosures are not supported by Potential for the first iteration of Piloting.
+      // We need to check inside disclosures or inside credential payload. Example path: "$.given_name"
+      let [matchedPath, matchedValue] = findMatchedClaim(
+        field.path,
+        disclosuresAsPayload
+      );
```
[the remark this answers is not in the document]
manuraf: here is more complicated using const because matchedPath really change...maybe enforce to use const could cause the code to become less readable

src/credential/presentation/07-evaluate-input-descriptor.ts
```diff
@@ -0,0 +1,235 @@
...
+  const requiredDisclosures: DisclosureWithEncoded[] = [];
+  const optionalDisclosures: DisclosureWithEncoded[] = [];
+
+  disclosures.forEach((disclosure) => {
+    if (requiredClaimNames.includes(disclosure.decoded[1]))
+      requiredDisclosures.push(disclosure);
+    else if (optionalClaimNames.includes(disclosure.decoded[1]))
+      optionalDisclosures.push(disclosure);
+    else if (isNotLimitDisclosure) optionalDisclosures.push(disclosure);
+  });
```
LazyAfternoons, line 122: Can we use the `filter` function here to make this more readable?

```suggestion
   const requiredDisclosures = disclosures.filter(
    (disclosure) => requiredClaimNames.includes(disclosure.decoded[1])
  );

  const optionalDisclosures = disclosures.filter(
    (disclosure) =>
      optionalClaimNames.includes(disclosure.decoded[1]) ||
      (isNotLimitDisclosure && !requiredClaimNames.includes(disclosure.decoded[1]))
  );
```

LazyAfternoons, line 122: If we value immutability over performance here, this could be replaced with two filter functions. I know this isn't enforced in this library as it is on other IO repositories so I'll leave this to you. It's just a suggestion.

LazyAfternoons (approved):
LGTM

PR #175, opened 2025-01-21
chore: [SIW-1268] Deletion of optional attributes from the credential issuer entity configuration 

RiccardoMolinari95:
#### List of Changes

This PR eliminates the attributes that aren't passed from the credential issuer entity configuration after verifying the technical specifications. Specifically, it removes optional attributes that are not specified in the technical rules.
In other cases, the optional attribute has been removed because the attribute is being passed.

Since these attributes are neither passed in the entity configuration nor specified in the technical rules, they can be safely removed.

#### Motivation and Context

This change is required to ensure that the entity configuration we receive is consistent with the one we subsequently parse

You should be able to obtain PID and credentials

#### Checklist:


mastro993 (approved):
LGTM

PR #181, opened 2025-01-29
feat: [WLEO-228] Enhance Request Object Handling and JWKS Retrieval using x5c

manuraf:
#### List of Changes

- Updated `types.ts `to include an optional `client_metadata` with `jwks_uri` and `jwks` fields for additional key information.
- Implemented certificate X.509 parsing and public key extraction in `04-retrieve-rp-jwks.ts` using `jsrsasign` and enhanced tests to cover new JWKS retrieval logic.

#### Motivation and Context

This PR aims to support X.509 certificates enables integration with systems that utilize certificate-based authentication.

#### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,10 +108,12 @@
     ]
   },
   "dependencies": {
+    "@types/jsrsasign": "^10.5.15",
     "ajv": "^8.17.1",
     "js-base64": "^3.7.7",
     "js-sha256": "^0.9.0",
     "jsonpath-plus": "^10.2.0",
+    "jsrsasign": "^11.1.0",
     "parse-url": "^9.2.0",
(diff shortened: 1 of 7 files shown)


LazyAfternoons (commented):
Should we move all the utility functions in the 4th step into our utility files according to their scope? Does it make sense to you?

manuraf:
> Should we move all the utility functions in the 4th step into our utility files according to their scope? Does it make sense to you?

manuraf:
it makes sense....in the future we could also move to [io-react-native-crypto](https://github.com/pagopa/io-react-native-crypto) as utility

LazyAfternoons (approved):
LGTM

PR #183, opened 2025-01-31
feat: [WLEO-218] Introduce unrequestedDisclosures for enhanced disclosure categorization

manuraf:
#### List of Changes

During evaluate-input-descriptor phase, introduced a new property `unrequestedDisclosures` to the EvaluatedDisclosures type to categorize disclosures that are neither required nor optional. When limit_disclosure is set to `required`, unrequestedDisclosures is empty, otherwise it contains disclosures that are not requested inside presentation_definition.

#### Motivation and Context

The introduction of `unrequestedDisclosures` addresses the need for a more granular categorization of disclosures in the credential presentation process. Previously, disclosures were limited to being either required or optional, which did not account for scenarios where certain disclosures should neither be mandatory nor suggested.

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #190, opened 2025-02-13
fix: Align authorization_details to OID4VCI spec

manuraf:
#### List of Changes

Change authorization_details schema with array format

#### Motivation and Context

This PR aims to fix a new issues about obtaining PID process. It was caused because authorization_details did not follow OID4VCI spec.

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #194, opened 2025-02-17
feat: [WLEO-263] Remote Presentation of multiple credentials and mdoc Support

manuraf:
#### List of Changes

* Introduced `evaluateInputDescriptors` and `prepareRemotePresentations` functions to handle multiple credentials and multiple input descriptors.
* Extended `08-send-authorization-response.ts` to properly manage single or multiple presentations (returning either a single vp_token or an array of tokens).
* Added preliminary support for `mso_mdoc` credentials with a placeholder implementation.
* Refactored example app code (presentation.ts) to demonstrate multi-credential handling.
* Refactor to obtain credential as base64 format rather than hex format.

#### Motivation and Context

Prior to this PR, the presentation flow supported only a single credential in the `vc+sd-jwt` format. This change enables the wallet to handle multiple credentials (including `mso_mdoc`) when generating and sending presentation tokens.

#### Checklist:


example/src/thunks/presentation.ts
```diff
@@ -78,43 +74,59 @@ export const remoteCrossDevicePresentationThunk = createAppAsyncThunk<
...
 
-  const disclosuresRequestedClaimName = [
-    ...requiredDisclosures.map((item) => item.decoded[1]),
-  ];
+  const credentialAndInputDescriptor = evaluateInputDescriptors.map(
+    (evaluateInputDescriptor) => {
+      // Present only the mandatory claims
+      const requestedClaims = [
+        ...evaluateInputDescriptor.evaluatedDisclosure.requiredDisclosures.map(
+          (item) => item.decoded[1]
```
grausof, line 104: I would add a comment why the second element of the array. I think because the first is the keytag

manuraf: the first is the claim name...in this case we prepare an array of claims which we want to share. These claims will be used to build the VpToken with disclosure....in real case, user will select the claims using  UI app

PR #195, opened 2025-02-27
feat: [SIW-1967] Align trust chain implementation with 0.9.0

ale-mazz:
#### List of Changes

- Added a new function, `buildTrustChain,` to build a Trust Chain, which will then be verified with the `verifyTrustChain` function.
- Fixed comments
- Aligned Federation types with 0.9.0

> [!NOTE]  
> x.509 certificates validation will be added later inside another PR.

#### Motivation and Context

This PR implements the changes needed for the alignment with [0.9.0 technical specifications](https://italia.github.io/eudi-wallet-it-docs/v0.9.0/en/trust.html)

> [!NOTE]  
> `buildTrustChain` cannot be tested until the provider releases a valid Trust Anchor.

#### Checklist:


src/trust/index.ts
```diff
@@ -259,3 +262,133 @@ export async function getSignedEntityStatement(
...
+  } = {}
+): Promise<string[]> {
+  return await appFetch(federationListEndpoint, {
+    method: "GET",
+  })
+    .then(hasStatusOrThrow(200))
+    .then((res) => res.text())
+    .then((text) => {
+      try {
+        return JSON.parse(text) as string[];
```
gispada, line 289: Instead of having a `try/catch` within a `then` block, why not using `Zod` with `safeParse`?

```suggestion
    const federationList = await fetch().then(FederationListResponse.safeParse);
    if (!federationList.success) // throw error
```

ale-mazz: Thanks, @gispada. I am still not used to Zod, and I forgot to add the validation even if it is a simple type.

src/trust/chain.ts
```diff
@@ -114,38 +85,51 @@ export async function validateTrustChain(
...
+
+      const entityStatementResult = EntityStatement.safeParse(decoded);
+      const entityConfigurationResult = EntityConfiguration.safeParse(decoded);
+
+      if (entityConfigurationResult.success) {
+        return getSignedEntityConfiguration(
+          entityConfigurationResult.data.payload.iss,
+          { appFetch }
+        );
+      } else if (entityStatementResult.success) {
```
gispada, line 107: Since we are returning early we can remove the `else if` and just use `if`, what do you think?

grausof (approved):
It seems ok for now. Given the complexity of the code, it should be tried on a test service. But what do you think of adding unit tests perhaps using non -regulatory examples?

ale-mazz:
> It seems ok for now. Given the complexity of the code, it should be tried on a test service. But what do you think of adding unit tests perhaps using non -regulatory examples?

Yes its a great idea, working on them right now. I'll update this PR with those tests!

ale-mazz:
@grausof So far I have
- Refactored `buildTrustChain` since it was not acting correctly
- I built some tests and mocked the HTTP response; everything should be aligned with what we expect.
- Fixed `zod` types
- Fixed mocks that weren't aligned with the spec.

Let me know if everything looks okay. Thanks!

grausof (approved):

grausof:
Great 🎉

PR #196, opened 2025-02-28
fix:[WLEO-277] State must be optional in Request Object and Presentation Response

manuraf:
#### List of Changes

- `state` claim is now optional in Request Object schema
- `state` will be added in remote presentation if it is present in  Request Object

#### Motivation and Context

This PR aims to fix the mandatory of `state` claim in  Request Object. [OID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-authorization-request) and I[talian Specs](https://italia.github.io/eudi-wallet-it-docs/v0.9.2/en/relying-party-solution.html#authorization-request-details) mark it as optional or recommended, not mandatory.

During Interop Test in Utrecht, a Relying Party did not include state inside its Request Object. You can try to test when it is present at https://verifier.eudiw.dev/.

#### Checklist:

### DIFF
--- a/src/credential/presentation/08-send-authorization-response.ts
+++ b/src/credential/presentation/08-send-authorization-response.ts
@@ -60,7 +60,7 @@ export const buildDirectPostBody = async (
   payload: DirectAuthorizationBodyPayload
 ): Promise<string> => {
   const formUrlEncodedBody = new URLSearchParams({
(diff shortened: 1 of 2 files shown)


LazyAfternoons (approved):
LGTM

PR #203, opened 2025-03-17
feat: [SIW-2106] Add remote presentation in the example app

gispada:
#### List of Changes

- Added the Presentation section to the example app to test the remote presentation flow.
- Minor changes to the presentation logic.
- More tests for DCQL queries.

#### Motivation and Context

This PR makes it easier to test remote presentation with the example app. It currently supports both `presentation_definition` and `dcql_query`.

> [!NOTE]
> The two modes of requesting a Verifiable Presentation are slightly different on the consumer side. We should decide whether to drop support for `presentation_definition` or to hide the two implementations behind a common interface.

Tested against PagoPA mock Relying Party and an institutional Relying Party.

#### Checklist:


grausof (approved):
Just a few small remarks and I would say to update the `Podfile.lock`

grausof (approved):
Great 👏🏻

PR #205, opened 2025-03-26
fix(example): re-add deep link handler for ios

ale-mazz:
#### List of Changes
Added the native function to handle deep links since it was removed and CIE ID L2 flow is not working anymore

Test a complete flow using CIE ID auth on the PID request

#### Checklist:


LazyAfternoons (approved):
LGTM

PR #206, opened 2025-03-26
feat: [WLEO-323] Add logging feature

LazyAfternoons:
> [!NOTE]
> Most of the changes are just related to logs. Most of the work is under `src/utils/logging.ts` and `example/src/utils/logging.ts`

#### List of Changes
- Add a `Logger` class under `src/utils/logging.ts`. This class is a singleton class which can be used to log data with a custom logger defined as required by `LogginContext`. This class is used in our flows to log useful data via the `log` method by providing the related log level. If no `LoggingContext` is provided then logging won't be performed;
- Add logging calls in existing flows;
- Add an implementation based on `react-native-logs` in our example app which logs via console and to a web server under `example/src/utils/logging.ts`;
- Add a debug server example in `expressjs` which consists of a single endpoint which receives the log and prints it in the home page;
- Add a section in the example app settings to set the logging server address.
- Update documentation;

#### Motivation and Context
This PR adds a logging system to the library and the example app as well. This can be useful while debugging the example app or any other app which uses this library. 
It also adds a very simple web server which can receive said logs and prints them in the homepage.
Feedback welcome on both implementation and other stuff which you think might be worth logging.

This can be tested by running: 
```bash
yarn
cd debug-logging-server
yarn
yarn node app.js 
cd .. 
cd example 
yarn 
cd ios && bundler exec pod install && cd .. 
yarn react-native run-ios || yarn react-native run-android
```
> [!WARNING]
> On Android you must use adb reverse tcp:XXXX tcp:XXXX where XXXX is the logging server port. On iOS if you're using a real device then you must use the address of your machine instead of `localhost` when typing it in the app's settings.

Open the web server address in the console in your browser.
Then run the example app and input the logging server address in the example app settings section by appending the `/sendLogs` endpoint (http://localhost:3045/sendLogs for example)

#### Checklist:


mastro993 (approved):
Great job!

PR #209, opened 2025-03-31
feat: [WLEO-266] Linked `io-react-native-cbor` bridge functions to verify and parse mDoc credentials and expanded iOS minimum supported version

sebastianopriscan:
This PR refines the credential issuance verification step by checking the obtained mDoc signature with the certificate present inside of the mDoc itself and the presence of the Wallet Instance issued key inside of the credential itself (though this last feature is now commented out for technical issues, the commented out lines have proper explanations that accompany them).

#### List of Changes

- `package.json`, `example/package.json`, `yarn.lock`, `example/yarn.lock` : upgraded `io-react-native-cbor` dependencies to 1.1.0.

- `example/ios/Podfile(.lock)` : changed deployment target and platform to minimum ios version supported.

- `src/utils/crypto.ts` : added a new utility method, `compareKeysByThumbprint`, that compares two JWKs by checking whether their thumbprint is equal.

- `src/mdoc/index.ts` : used `io-react-native-cbor`'s exposed bridge methods and utilities mentioned above to extract the `issuerSigned` from a raw mDoc, get the signing key from its certificate and check the mDoc's signature.

- `src/credential/issuance/07...credential.ts` : changed `verifyandParseCredential`'s signature to accept a `credentialType` parameter, added for now commented out "deviceKey in credential" presence check.

Other changes are related to the just mentioned ones.

#### Motivation and Context

This PR solves issue [\[WLEO-266\]](https://pagopa.atlassian.net/jira/software/c/projects/WLEO/boards/601?assignee=712020%3Adcf931c4-2a77-403b-b539-7a018b6d504c&selectedIssue=WLEO-266).

0. Due to changes to dependencies, start from a clean build on Android and iOS.
    ```bash
    cd android && ./gradlew clean && cd ..
    cd ios && rm -rf Pods/ && rm -f Podfile.lock && bundle exec pod install && cd ..
    ```
1. Setup the Wallet Instance and obtain a PID.
2. Try and obtain an mDL multiple times, **every** attempt **should** work.

#### Checklist:


manuraf (approved):

src/utils/crypto.ts
```diff
@@ -106,3 +107,49 @@ export const getSigningJwk = (publicKey: RSAKey | KJUR.crypto.ECDSA): JWK => ({
...
+export const deepCompareKeys = (key1: PublicKey, key2: PublicKey) => {
+  if (key1.kty === "EC" && key2.kty === "EC") {
+    return key1.crv === key2.crv && key1.x === key2.x && key1.y === key2.y;
+  }
+  if (key1.kty === "RSA" && key2.kty === "RSA") {
+    return key1.alg === key2.alg && key1.e === key2.e && key1.n === key2.n;
+  }
+
+  return false;
+};
```
grausof, line 130: This comparison is usually not recommended. The best way to verify that two keys are the same is to calculate the thumbprint of both and compare it. The methods for calculating the thumbprint are already present here.

grausof (changes requested):

example/ios/Podfile
```diff
@@ -31,7 +31,7 @@ target 'IoReactNativeWalletExample' do
 
   post_install do |installer|
     # TODO: can be removed after upgrading RN version [WLEO-170]
-    deployment_target = '13.0'
+    deployment_target = min_ios_version_supported
     installer.generated_projects.each do |project|
         project.targets.each do |target|
             target.build_configurations.each do |config|
```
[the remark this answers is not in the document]
manuraf: Removing `deployment_target = ..` and `installer.generated_projects.each do` builds and works correctly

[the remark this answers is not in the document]
sebastianopriscan: Commit 7ef984c should have solved this issue.

src/utils/crypto.ts
```diff
@@ -106,3 +107,45 @@ export const getSigningJwk = (publicKey: RSAKey | KJUR.crypto.ECDSA): JWK => ({
...
+ * @param key1 The first key
+ * @param key2 The second key
+ * @returns true if the keys are deep equal, false otherwise
+ */
+export const deepCompareKeys = async (key1: PublicKey, key2: PublicKey) => {
+  const thumbprint1 = await thumbprint(key1);
+  const thumbprint2 = await thumbprint(key2);
+
+  return thumbprint1 === thumbprint2;
+};
```
LazyAfternoons, line 126: I would change the name of this function as we are not deeply comparing something, just their thumbprint. Also the documentation should be updated as well. Does it make sense to you?

LazyAfternoons (dismissed):

src/utils/crypto.ts
sebastianopriscan: You're right, I'll fix soon

src/utils/crypto.ts
```diff
@@ -106,3 +107,44 @@ export const getSigningJwk = (publicKey: RSAKey | KJUR.crypto.ECDSA): JWK => ({
...
+  key2: PublicKey
+) => {
+  const thumbprint1 = await thumbprint(key1);
+  const thumbprint2 = await thumbprint(key2);
+
+  return thumbprint1 === thumbprint2;
+};
+
+/**
+ * This helper function converts a DEM certificate in PEM format by adding newlines and
```
grausof, line 128:
```suggestion
 * This helper function converts a DER certificate in PEM format by adding newlines and
```

src/utils/crypto.ts
```diff
@@ -106,3 +107,44 @@ export const getSigningJwk = (publicKey: RSAKey | KJUR.crypto.ECDSA): JWK => ({
...
+export const derToPem = (der: string | ArrayBuffer): string => {
+  let base64: string;
+
+  // Se 'der' è un ArrayBuffer, convertiamolo in una stringa Base64
+  if (der instanceof ArrayBuffer) {
+    // Converte l'ArrayBuffer in una stringa Base64
+    base64 = Buffer.from(new Uint8Array(der)).toString("base64");
+  } else {
+    // Se è già una stringa, assumiamo che sia in Base64
+    base64 = der;
```
grausof, line 142: Better all comments in English

grausof, line 142: Why this assumption?

src/utils/crypto.ts
```diff
@@ -106,3 +107,44 @@ export const getSigningJwk = (publicKey: RSAKey | KJUR.crypto.ECDSA): JWK => ({
...
+ * @returns true if the keys' thumbprints are equal, false otherwise
+ */
+export const compareKeysByThumbprint = async (
+  key1: PublicKey,
+  key2: PublicKey
+) => {
+  const thumbprint1 = await thumbprint(key1);
+  const thumbprint2 = await thumbprint(key2);
+
+  return thumbprint1 === thumbprint2;
```
grausof, line 124: what do you think about using promise.all?

sebastianopriscan: I agree, I'll publish a fix soon

src/utils/crypto.ts
sebastianopriscan: It was a quick implementation, but now that I think about it, what would be the best thing to do? Should I just specify that the method *should* take only base64 encoded strings, and leave to the programmer the responsibility of passing a correct string, or should I perform a check on the string and throw an exception in case it's not correct?

LazyAfternoons: In my opinion you should specify that this method takes base64 encoded strings. You can also validate that by using `js-base64` which provides a method to check for its validity. 

On a side node, we strongly support immutable code, thus the usage of `let` is discouraged even though not enforced through linting rules in this repository.

Let me know what you think.

sebastianopriscan: I agree with you, but by looking at the code, I found another function in the same file, `convertCertToPem`, which, line breaks aside, does pretty much the same thing but just takes in a string. Being that, from our use cases, the DER we receive is already a base64 encoded string, and being that `convertCertToPem` doesn't do checks that are that strict, couldn't the `derToPem` method be merged in `convertCertToPem`, removing type checking, conversions and `ByteArray` support, and keeping just the line breaks?

Please let me know wheter you agree with this or not.

src/mdoc/index.ts
```diff
@@ -1,28 +1,36 @@
...
+  if (!issuerSigned) {
     throw new Error("Invalid mDoc");
   }
 
-  const sigKey = Array.isArray(publicKey)
-    ? publicKey.find((k) => k.use === "sig")
-    : publicKey;
-  sigKey;
+  const cert = issuerSigned.issuerAuth.unprotectedHeader[0]?.keyId;
+  if (!cert) throw new Error("Certificate not present in credential");
```
grausof, line 18: Why is it called a certificate? Kid is the id of a key

src/mdoc/index.ts
```diff
@@ -1,28 +1,36 @@
...
+
+  const pemcert = derToPem(b64utob64(cert));
+  const publickey = parsePublicKey(pemcert);
+  if (!publickey) throw new Error("Certificate not present in credential");
```
grausof, line 22: ditto. Maybe I don't have the right context but looking at the code it seems that it is only converting the key id. Are you sure it has to work this way? I suggest adding more unit-tests on this

grausof (changes requested):

src/mdoc/index.ts
manuraf: `keyId` should be `cert`, the problem is related to `iso18013-android` sdk at [this](https://github.com/pagopa/iso18013-android/blob/1ea5a220f16cafb5b077d6c778519c536722015d/cbor/src/main/java/it/pagopa/io/wallet/cbor/model/ModelMDoc.kt#L310) point and in `io-react-native-cbor` as consequence ....I propose to open an issue there and fix it in the next version

src/utils/crypto.ts
sebastianopriscan: Commit 09f6dbe should have solved this issue

src/utils/crypto.ts
sebastianopriscan: Commit https://github.com/pagopa/io-react-native-wallet/commit/09f6dbe89b8d9952d7a51bf3874d6804ed64f0f9 should have solved this issue.

src/utils/crypto.ts
sebastianopriscan: Commit 63ca980 should have solved this issue.

src/utils/crypto.ts
manuraf: I think that you could use `convertCertToPem` and removing `derToPem`....in this context the certificate seems to be in base64 format

sebastianopriscan: Replaced and removed `derToPem` with commit 5f61fe2

grausof (approved):

PR #210, opened 2025-04-01
feat: [SIW-2196] Update ObtainCredential for IT Wallet transition

RiccardoMolinari95:
#### List of Changes

- Added an optional parameter `operationType` to the `obtainCredential` method to distinguish the issuing phase
- Added `operationType` to the `header` in the `/credential` request for the `re-issuing` flow

#### Motivation and Context

This PR aims to update the `obtainCredential` method for the transition from Documenti su IO to IT Wallet.

It is essential to distinguish between a standard `issuing` request and a `re-issuing` request.

It is currently not possible to test the credential flow in the re-issuing mode

#### Checklist:

### DIFF
--- a/src/credential/issuance/06-obtain-credential.ts
+++ b/src/credential/issuance/06-obtain-credential.ts
@@ -27,7 +27,8 @@ export type ObtainCredential = (
     dPopCryptoContext: CryptoContext;
     credentialCryptoContext: CryptoContext;
     appFetch?: GlobalFetch["fetch"];
-  }
+  },
+  operationType?: "reissuing"
 ) => Promise<CredentialResponse>;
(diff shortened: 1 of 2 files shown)


mastro993:
What do you think about adding test to ensure the correct behaviour of the updated function?

RiccardoMolinari95:
> What do you think about adding test to ensure the correct behaviour of the updated function?

Yes, you're right! Added tests in [df95ae0](https://github.com/pagopa/io-react-native-wallet/pull/210/commits/df95ae0e0f4e0783702e70f6cccf2ffcc9541e06)

grausof (approved):
LGTM

PR #211, opened 2025-04-01
fix: [WLEO-339] Add function to remove leading zero from base64url encoded JWK properties

manuraf:
> [!WARNING]
> Breaking change for IT Wallet!

#### List of Changes

- New helper function, `fixBase64WithLeadingZero` (with its underlying helper, `removeLeadingZeroAndParseB64Url`), to sanitize JSON Web Key (JWK) properties. It addresses potential issues with cryptographic operations by processing the base64url encoded key components (x, y, e, and n). The new functions decode the input, remove any unwanted leading zero, and then re-encode the result to ensure proper formatting. 
- It replace `fixBase64EncodingOnKey` inside `CryptoContext`

#### Motivation and Context

The method `getPublicKey` in the `io-react-native-crypto` library returns a key with coordinates in base64 format. This base64 format is generated by converting hex to base64 internally, but sometimes the conversion does not conform to the P-256 key standard—which requires exactly 32 bytes—because the hex representation may result in 33 bytes.

The problem stems from the behavior of BigInteger.toByteArray, which, in certain contexts (e.g., ASN.1 DER signatures), prepends a 0x00 byte if the coordinate starts with a value greater than 0x7F to avoid it being interpreted as a negative number. 
This PR aims to temporarily bypass the issue within this library by ensuring that any unwanted leading zero is removed during the conversion process.

- Re-issuance Wallet Instance and PID
- Try a PID Presentation, it should work as the same

#### Checklist:

### DIFF
--- a/src/utils/crypto.ts
+++ b/src/utils/crypto.ts
@@ -27,7 +27,7 @@ export const createCryptoContextFor = (keytag: string): CryptoContext => {
      */
     async getPublicKey() {
       return getPublicKey(keytag)
-        .then(fixBase64WithLendingZero)
+        .then(fixBase64WithLeadingZero)
         .then(async (jwk) => ({
           ...jwk,
           // Keys in the TEE are not stored with their KID, which is supposed to be assigned when they are included in JWK sets.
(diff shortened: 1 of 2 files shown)


src/utils/crypto.ts
```diff
@@ -48,6 +49,45 @@ export const createCryptoContextFor = (keytag: string): CryptoContext => {
...
 };
 
+/**
+ * This function takes a JSON Web Key (JWK) and returns a new JWK with its base64-url properties (x, y, e, n) processed.
+ * Each property is passed through the `removeLeadingZeroAndParseb64u` function if it exists, which fixes any unwanted leading zeros.
+ *
+ * @param key - The input JSON Web Key that may contain properties with potential leading zero issues.
+ * @returns A new JSON Web Key with the processed properties.
+ */
+const fixBase64WithLeadingZero = (key: JWK): JWK => {
```
grausof, line 59: I think it's useful to add some unit tests

grausof (changes requested):

src/utils/crypto.ts
manuraf: addressed with [2ef5e67](https://github.com/pagopa/io-react-native-wallet/pull/211/commits/2ef5e676a63492ab5ebcfd1ad89ccf02c745c81b)

PR #215, opened 2025-04-08
feat: [WLEO-345] Prioritize EC keys first P256 during Remote Presentation enc

manuraf:
#### List of Changes

- Updated `choosePublicKeyToEncrypt` function to prioritize **EC** keys (P-256, P-384) over **RSA** keys when selecting encryption keys from the JWKS.
- Filtered keys by use: "enc" and sorted them by type and curve support.
- Added a new unit test to verify that the function selects the first supported EC key.

#### Motivation and Context

This change ensures compatibility with `io-react-native-jwt`, which guarantees supports only for EC encryption keys with curves `P-256` and `P-384`. The previous implementation defaulted to the first enc key without prioritizing supported types, which could lead to runtime errors or unexpected behavior. By explicitly preferring compatible EC keys, we improve reliability and interoperability with supported clients.

#### Checklist:

### DIFF
--- a/src/credential/presentation/08-send-authorization-response.ts
+++ b/src/credential/presentation/08-send-authorization-response.ts
@@ -30,6 +30,9 @@ export const AuthorizationResponse = z.object({
  * Selects a public key (with `use = enc`) from the set of JWK keys
  * offered by the Relying Party (RP) for encryption.
  *
+ * Preference is given to EC keys (P-256 or P-384), followed by RSA keys,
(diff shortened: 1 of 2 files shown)


sebastianopriscan:
I think that, after these proposed updates, the javadoc for the `choosePublicKeyToEncrypt` method and the comment on its invocation on line `113` of `src/credential/presentation/08-send-authorization-response.ts` should be expanded to reflect the new policies introduced, let me know if you agree with that or not.

PR #216, opened 2025-04-09
chore: [SIW-2230] Expose credentials not found in DCQL presentation

gispada:
#### List of Changes

- Rename `CredentialNotFoundError` to `CredentialsNotFoundError`
- Throw `CredentialsNotFoundError` when the requested credentials could not be found
- Update DCQL queries tests

#### Motivation and Context

This PR modifies the `CredentialNotFoundError` to include details on credentials that are requested for presentation but could not be found. The goal is to provide useful information to a consumer that wants to display the missing credentials.

Start a presentation that requires a credential not found in the wallet. Check the debug info to see `CredentialNotFoundError`.

#### Checklist:


mastro993 (approved):
Everything works as expected on Android, but I have issues with the QR scanner screen on iOS: the app crashes as soon as I navigate to it. 
With some debugging I found that the error comes from the permission request function.
The fix is out of scope for this PR, so if you prefer we can merge this and think about it in a separate one.

PR #229, opened 2025-05-16
fix: [WLEO-406] Updated sd-jwt and mdoc credential parsers

sebastianopriscan:
This PR updates `sd-jwt` and `mdoc` credential parsers to exclude non mandatory claims that are not present in the credentials and throw errors in case mandatory claims are not present.

#### List of Changes

- Added filtering in `parseCredentialSdJwt` and `parseCredentialMDoc` to exclude claims defined in the `issuerConfig`s as non mandatory and not present in the obtained credential
- Added check in `parseCredentialMDoc` that throws an error when a credential doesn't contain mandatory, claims, adapted from `parseCredentialSdJwt`'s own check

#### Motivation and Context

This PR solves an issue for which some credentials' claims were not rendered correctly because they were only defined in the `issuerConf`, and not in the credential itself.

The new functionalities have been tested with specific unit tests whose passing is a necessary condition for merging this PR.

#### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,9 @@
     ],
     "transformIgnorePatterns": [
       "node_modules/(?!(jest-)?@react-native|react-native|uuid|@pagopa/io-react-native-cbor)"
+    ],
+    "testPathIgnorePatterns": [
+      "/__tests__/utils"
     ]
   },
   "react-native-builder-bob": {
--- a/src/credential/issuance/07-verify-and-parse-credential.ts
+++ b/src/credential/issuance/07-verify-and-parse-credential.ts
@@ -64,7 +64,7 @@ type DecodedMDocCredential = Out<typeof verifyMdoc> & {
   issuerSigned: CBOR.IssuerSigned;
 };
 
-const parseCredentialSdJwt = (
+export const parseCredentialSdJwt = (
   // the list of supported credentials, as defined in the issuer configuration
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -96,6 +96,9 @@
     ],
     "transformIgnorePatterns": [
       "node_modules/(?!(jest-)?@react-native|react-native|uuid|@pagopa/io-react-native-cbor)"
+    ],
+    "testPathIgnorePatterns": [
+      "/__tests__/utils"
```
LazyAfternoons, line 101: Why is this change necessary?

sebastianopriscan: I added that line because otherwise jest would consider the files in the `utils` folder as test files and throw errors, and I added utilities for a test in a folder with such directory schema

sebastianopriscan: More precisely, I added the tests inside the `src/credential/issuance/__tests__/utils/` directory.

LazyAfternoons: I suggest moving the utils out of `__tests__` to avoid this. I think they fall under `utils`.

sebastianopriscan: With commit [08e7822](https://github.com/pagopa/io-react-native-wallet/pull/229/commits/08e7822f862cb1a2405f98383661423182e27df2) the utils have been moved

manuraf (approved):

PR #233, opened 2025-05-29
feat: [SIW-2094] Add trust federation complete flow in example app

ale-mazz:
#### List of Changes

This PR adds the Trust Federation verification flow inside the example app, including the x509 validation from `io-react-native-crypto`
#### Motivation and Context

- Bumped `io-react-native-crypto` to `v1.2.1`
- Added x509 verification flow
- Fixed some bugs in the trust flow
- Added the new PRE-PROD trust anchor endpoint
- Added the new Trust Federation screen with connected logic, with three example flows in the example app
- Added a README about the Trust Federation verification process

Run the trust federation flow using the endpoints provided in the example app. You can add more tests with different federated entities.

#### Checklist:

### DIFF
--- a/src/trust/chain.ts
+++ b/src/trust/chain.ts
@@ -134,22 +134,18 @@ export async function validateTrustChain(
       );
     }
 
-    const originalX5cChain: string[] = jwkUsedForVerification.x5c;
-    let certChainToPassToNative = [...originalX5cChain];
-
     // If the chain has more than one certificate AND
     // the last certificate in the x5c chain is the same as the trust anchor,
     // remove the anchor from the chain being passed, as it's supplied separately.
-    if (
-      certChainToPassToNative.length > 1 &&
-      certChainToPassToNative.at(-1) === x509TrustAnchorCertBase64
-    ) {
-      certChainToPassToNative.pop(); // Remove the last element
-    }
+    const certChainBase64 =
+      jwkUsedForVerification.x5c.length > 1 &&
+      jwkUsedForVerification.x5c.at(-1) === x509TrustAnchorCertBase64
+        ? jwkUsedForVerification.x5c.slice(0, -1)
+        : jwkUsedForVerification.x5c;
 
     const x509ValidationResult: CertificateValidationResult =
       await verifyCertificateChain(
-        certChainToPassToNative,
+        certChainBase64,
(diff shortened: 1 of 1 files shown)


src/trust/chain.ts
```diff
@@ -82,13 +96,73 @@ export async function validateTrustChain(
...
+      const originalX5cChain: string[] = jwkUsedForVerification.x5c;
+      let certChainToPassToNative = [...originalX5cChain];
+
+      // If the chain has more than one certificate AND
+      // the last certificate in the x5c chain is the same as the trust anchor,
+      // remove the anchor from the chain being passed, as it's supplied separately.
+      if (
+        certChainToPassToNative.length > 1 &&
+        certChainToPassToNative[certChainToPassToNative.length - 1] ===
+          x509TrustAnchorCertBase64
```
adelloste, line 138:
```suggestion
        certChainToPassToNative.at(-1) === x509TrustAnchorCertBase64
```

ale-mazz:
> Tested on iOS physical device and everything works fine (both for PagoPA WP and RP). Just noticed that `Podfile.lock` changes after `bundle exec pod install`. Do you have the same behavior on your side?

No, I do not have that behavior. Could it be your Cocoapods version?

ale-mazz:
Thanks @adelloste, everything should be fixed!

src/trust/chain.ts
```diff
@@ -82,13 +97,78 @@ export async function validateTrustChain(
...
+
+    // If the chain has more than one certificate AND
+    // the last certificate in the x5c chain is the same as the trust anchor,
+    // remove the anchor from the chain being passed, as it's supplied separately.
+    if (
+      certChainToPassToNative.length > 1 &&
+      certChainToPassToNative.at(-1) === x509TrustAnchorCertBase64
+    ) {
+      certChainToPassToNative.pop(); // Remove the last element
+    }
```
adelloste, line 148: What do you think about doing something like this?
```suggestion
const certChainBase64 =
   jwkUsedForVerification.x5c.length > 1 &&
   jwkUsedForVerification.x5c.at(-1) === x509TrustAnchorCertBase64
     ? jwkUsedForVerification.x5c.slice(0, -1)
     : jwkUsedForVerification.x5c;
```

PR #239, opened 2025-06-10
fix: [WLEO-365] Publish app workflow path and and other minor fixes

LazyAfternoons:
#### List of Changes
- Move the workflow in the root directory as it won't be detected in subfolders;
- Fix the environment variable syntax; 
- Add GH token permissions;
- Update ruby setup.

#### Motivation and Context
This PR fixes the android build in artifacts workflow along with other minor fixes.

#### Checklist:


mastro993 (approved):
LGTM

PR #242, opened 2025-06-12
ci: create separate job to check release name and fix publish workflow

gispada:
#### Motivation and Context

This PR fixes a problem in the publish workflow when the release name is checked: the function `toLower` is not valid in GitHub expressions. Additionally, the if condition was at the single step level, not at the job level.

The PR adds a separate job to check the release name and uses [GitHub outputs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs) to pass the information to the build job.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,10 +14,10 @@ jobs:
         id: check_title
         run: |
           if echo "${{ github.event.release.name }}" | grep -iq 'example'; then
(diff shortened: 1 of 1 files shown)


LazyAfternoons (commented):
LGTM

gispada:
> Shouldn't `GITHUB_OUTPUT` be quoted?

Nice catch. Fixed in e11e9ae76197ed5da2b0c5e0b47ad2d3edea268e.

PR #248, opened 2025-07-01
chore: rename `credentialType` to `credentialId`

gispada:
#### List of Changes

- Renamed `credentialType` paramater to `credentialId` in the issuance functions

#### Motivation and Context

With the new credentials the credential type is no longer unique, as the same credential might be available in multiple formats. The issuance functions now require a credential configuration ID that will be requested to the Issuer; this ID can be mapped with one of the keys found in the Issuer's EC `credential_configurations_supported`.

To avoid ambiguities, `credentialType` has been renamed in every occurrence that requests a credential configuration ID.


mastro993 (approved):
LGTM

PR #261, opened 2025-07-24
feat: [SIW-2794] Update React Native to version 0.78.3

LazyAfternoons:
#### List of Changes
- Update the project according to the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.75.5&to=0.78.3);
- Update a few [dependencies](https://github.com/pagopa/io-react-native-wallet/pull/261/files#diff-a593fc14fdef41aaf092f774b92cb74bafeda3e9559f8ae4762374bf4af80c92) which were incompatibile with 0.78;
- Update `ruby` and `node` versions;
- Update a few components due to the new design system version;
- Update workflow to use the node version specified in the `.node-version` file.

#### Motivation and Context
This PR updates the React Native to version 0.78.3. This is needed in order to support `io-react-native-iso18013` library which is compatibile from 0.78.x onward. The new architecture is still disabled as there are some issues with modals on Android.

```bash
cd example/android && ./gradlew clean && cd ...
rm -rf node_modules 
rm -rf example/node_modules
rm -rf example/ios/DerivedData -> if saved locally
rm -rf example/ios/Pods 
```
Then follow the [README.md](https://github.com/pagopa/io-react-native-wallet/blob/master/example/README.md). Be sure to pick up the proper ruby and node version.
Test on both platforms that everything is working as before.

#### Checklist:


example/ios/AppDelegate.swift
ChrisMattew, line 1: Was this file added by mistake?

LazyAfternoons: Indeed it was, remove. Thank you.

ale-mazz:
Tested on iOS, everything appears to be working correctly without any regressions.
Tested on Android, and we have a regression in the PID flow, specifically in the CIE+PIN step. Even hardcoding the CIE pin, the modal is broken, and the process is stuck on loading.

ale-mazz (approved):
LGTM. Let's also wait for @ChrisMattew approval.

ChrisMattew (approved):
Tested on Android and it works as expected! 
LGTM 💪🏻

PR #265, opened 2025-07-29
feat: [SIW-2740] Add verify and parse mdoc

ChrisMattew:
#### List of Changes
- Added the `@pagopa/io-react-native-iso18013` and `jsrsasign ` packages
- Upgraded the `@pagopa/io-react-native-crypto` package
- Restricted the `format` type returned by `obtainCredential` to a union of string literals (`dc+sd-jwt` and `mso_mdoc`)
- Added the `x5chain` verification and the signature validation for the credentials in `mdoc` format
- Added the parsing function for credentials in `mdoc` format, preserving the namespace while keeping the structure of `ParsedCredential` unchanged compared to the `dc+sd-jwt` format
- Updated the example app and documentation accordingly, based on the changes made in the core library

#### Motivation and Context
This PR adds support for credentials in `mdoc` format, integrating the verification and parsing mechanisms for the credential

#### How to test it?
From the example app, once you have obtained what is needed to request credentials, request the MDL in `mdoc` format.
The credential verification must succeed (using a proxy tool, verify that after the `/credential` call, the `/pki/ta-sub.crl` endpoint is also called).
Once the credential has been obtained, using the debug interface available in the app, make sure that the structure of the `parsedCredential` matches the example below:
```javascript
{
 // ...
 "org.iso.18013.5.1:claim_name": {
    "value": "...",
    "name": {
        "it-IT": "...",
        "en-EN": "...",
     }
  }
 // ...
}
``` 
> [!NOTE]
> The test can be performed in both the PRE and PROD environments.

#### Demo:
| PRE | PROD |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/9d960500-781c-4e89-b8f6-f9232e310df0"></video> | <video src="https://github.com/user-attachments/assets/1f28508e-7f7a-4419-9ed8-ae5b4d01252c"></video>| 

#### Checklist:


src/mdoc/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  MissingX509CertsError,
+  X509ValidationError,
+} from "../trust/errors";
+import { getTrustAnchorEntityConfiguration } from "../trust/build-chain";
+import { IoWalletError } from "../utils/errors";
+
+export const verify = async (
+  token: string,
+  issuerKeys: JWK[],
+  authorityHints?: string[]
```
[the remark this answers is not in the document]
ChrisMattew: Yes, I'll replace them, thank you!

src/mdoc/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    throw new IoWalletError("Invalid mDoc");
+  }
+  if (!authorityHints?.length) {
+    throw new FederationError("Missing authorityHints");
+  }
+  if (!issuerSigned.issuerAuth.unprotectedHeader.x5chain) {
+    throw new MissingX509CertsError("Missing x509 certificates");
+  }
+
+  await verifyTrustChain(authorityHints, issuerSigned);
```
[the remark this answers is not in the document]
ChrisMattew: Do you mean the `x5chain`? 

**eg:**
```typescript
 await verifyCertificateChain(
          x5chain,
          X509CertificateRoot,
          options
 );
```

example/src/screens/ProximityScreen.tsx
```diff
@@ -0,0 +1,402 @@
...
+  const startFlow = useCallback(async () => {
+    const trustAnchorEntityConfig =
+      await Trust.Build.getTrustAnchorEntityConfiguration(WALLET_TA_BASE_URL, {
+        appFetch,
+      });
+
+    const x5c = trustAnchorEntityConfig.payload.jwks.keys[0]?.x5c;
+    if (!x5c) {
+      throw new Error("No suitable x5c found in Trust Anchor JWKS.");
+    }
```
grausof, line 277: Use `getTrustAnchorX509Certificate` from `example/src/utils/credential.ts`

src/utils/crypto.ts
```diff
@@ -55,3 +57,43 @@ export const withEphemeralKey = async <R>(
   const ephemeralContext = createCryptoContextFor(keytag);
   return fn(ephemeralContext).finally(() => deleteKey(keytag));
 };
+/**
+ * Converts a certificate string to PEM format.
+ *
+ * @param certificate - The certificate string.
+ * @returns The PEM-formatted certificate.
+ */
+export const convertCertToPem = (certificate: string): string =>
```
grausof, line 66: It's better to specify that this is a DER to PEM conversion. With the DER already encoded in base64

src/utils/crypto.ts
```diff
@@ -55,3 +57,43 @@ export const withEphemeralKey = async <R>(
...
+  const x509 = new X509();
+  x509.readCertPEM(pemCert);
+  const publicKey = x509.getPublicKey();
+
+  if (publicKey instanceof RSAKey || publicKey instanceof KJUR.crypto.ECDSA) {
+    return publicKey;
+  }
+
+  return undefined;
+};
```
grausof, line 88: To avoid the dependency on jsrsasign what do you think about using a JWK and therefore [PublicKey type](https://github.com/pagopa/io-react-native-crypto/blob/fed9420e0bb961a410642dc13718518e5aeda917/src/index.tsx#L74C13-L74C22) from @pagopa/io-react-native-crypto ?

src/credential/issuance/07-verify-and-parse-credential.ts
```diff
@@ -176,6 +299,57 @@ async function verifyCredentialSdJwt(
...
+async function verifyCredentialMDoc(
+  rawCredential: string,
+  x509CertRoot: string,
+  holderBindingContext: CryptoContext
+): Promise<DecodedMDocCredential> {
+  /**
+   * For the moment, being that issues in the crypto key generation
+   * have been found on Android, the check for the deviceKey inside
+   * of the mDoc is skipped, so we are not interested in the holderBindingKey
+   */
```
grausof, line 326: I would add a TODO with the issue id on Jira

src/credential/issuance/07-verify-and-parse-credential.ts
```diff
@@ -176,6 +299,57 @@ async function verifyCredentialSdJwt(
...
+   * have been found on Android, the check for the deviceKey inside
+   * of the mDoc is skipped.
+   */
+  //const key = decodedCredential.mDoc.issuerSigned.issuerAuth.payload.deviceKeyInfo.deviceKey;
+  //
+  //if (!compareKeysByThumbprint(key, holderBindingKey as PublicKey)) {
+  //  throw new IoWalletError(
+  //    `Failed to verify holder binding, holder binding key and mDoc deviceKey don't match`
+  //  );
+  //}
```
grausof, line 349: but the problem hasn't been fixed?

src/credential/issuance/07-verify-and-parse-credential.ts
ChrisMattew: I restored it, as I answered [here](https://github.com/pagopa/io-react-native-wallet/pull/265/files/c3606b62887b181410184f60abdfe08d5a094e75#r2247753681)

src/utils/crypto.ts
ChrisMattew: It can be used for this case, but `jsrsasign` cannot be removed for now, as it's still used in other scenarios, for example, in this file at line 80

src/utils/crypto.ts
ChrisMattew: I agree, I'll change it. Thank you

src/utils/crypto.ts
grausof: It's fine to use it internally within the function. I would just avoid using types from other packages like return .

src/utils/crypto.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-react-native-wallet/pull/265/commits/19102872b1369bac7e15e2aa4c57b0ad621d6cd8)

src/utils/crypto.ts
ChrisMattew: Ok, so I'll change it asap

grausof: What do you think about throwing an error instead of returning undefined?

ChrisMattew: Addressed [here](https://github.com/pagopa/io-react-native-wallet/pull/265/commits/6697ef63e63dda23e0b3b0971b7f210bb392ef1e). 
In the end, the `return undefined` was removed in favor of throwing an error if the `publicKey` is not supported.

grausof (approved):
Great work 🎉

src/utils/crypto.ts
```diff
@@ -55,3 +59,56 @@ export const withEphemeralKey = async <R>(
...
+  key1: PublicKey,
+  key2: PublicKey
+) => {
+  //Parallel for optimization
+  const [thumbprint1, thumbprint2] = await Promise.all([
+    thumbprint(key1),
+    thumbprint(key2),
+  ]);
+  return thumbprint1 === thumbprint2;
+};
```
grausof, line 114: You already have `isSameThumbprint` from `utils/jwks.ts`

PR #266, opened 2025-07-30
feat: [SIW-2523, SIW-2706] Add support for education_degree and education_enrollment

ale-mazz:
#### List of Changes
- Added two new credentials inside the Credential screen
- Modified the sd-jwt parser to support nested paths
- Added a new test for nested paths

#### Motivation and Context
This PR adds the support for `education_degree ` and `education_enrollment` credentials

- Try to obtain those two new credentials and verify the parsing is correct and complete (Be sure to use a PRE account with those credentials)

#### Checklist:

### DIFF
--- a/src/credential/issuance/07-verify-and-parse-credential.ts
+++ b/src/credential/issuance/07-verify-and-parse-credential.ts
@@ -285,16 +285,17 @@ const parseCredentialMDoc = (
   // and filter the non present ones
   const attrsNotInDisclosures = attrDefinitions.filter(
     ([attrDefNamespace, attrKey]) =>
-      flatNamespaces.some(
+      !flatNamespaces.some(
         ([namespace, claim]) =>
           attrDefNamespace === namespace && attrKey === claim
       )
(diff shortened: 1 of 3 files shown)


src/credential/issuance/07-verify-and-parse-credential.ts
```diff
@@ -73,55 +73,110 @@ const parseCredentialSdJwt = (
...
+  // Helper to build localized names
+  const buildName = (display: { locale: string; name: string }[]) =>
+    display.reduce(
+      (names, { locale, name }) => ({ ...names, [locale]: name }),
+      {} as Record<string, string>
+    );
+
+  // Recursive helper to build a nested object structure from a given path
+  const applyPath = (
+    target: any,
```
gispada, line 101: Do you think it is possible to remove these `any`'s? I would also try to improve the function/variable names, I find this code a bit hard to follow.

gispada (approved):
LGTM

PR #268, opened 2025-07-31
feat: [WLEO-537] Update React Native to version 0.78.3

LazyAfternoons:
> [!NOTE]
> This is a backport of #261 

#### List of Changes
- Update the project according to the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.75.5&to=0.78.3);
- Update a few dependencies which were incompatibile with 0.78;
- Update ruby and node versions;
- Update a few components due to the new design system version;
- Update workflow to use the node version specified in the .node-version file;
- Prevent multiple QR code scans in the scanner screen which caused an error due to `navigation.goBack` being called several times; 
- Remove the persistence of the presentation state. 

#### Motivation and Context
This PR updates the React Native to version 0.78.3. This is needed in order to support io-react-native-iso18013 library which is compatibile from 0.78.x onward. The new architecture is still disabled as there are some issues with modals on Android and with `@pago/io-react-native-login-utils` on iOS.

```bash
cd example/android && ./gradlew clean && cd ...
rm -rf node_modules 
rm -rf example/node_modules
rm -rf example/ios/DerivedData -> if saved locally
rm -rf example/ios/Pods 
```
Then follow the [README.md](https://github.com/pagopa/io-react-native-wallet/blob/master/example/README.md). Be sure to pick up the proper ruby and node version.
Test on both platforms that everything is working as before.

#### Checklist:


LazyAfternoons:
> I have encountered an error during remote presentations on an iOS 18.5 device, which seems to be related to a navigator problem, here the details
> 
> ```
> The action 'GO_BACK' was not handled by any navigator.
> 
> Is there any screen to go back to?
> 
> This is a development-only warning and won't be shown in production.
> ```
> 
> Below there's a stack trace of the error, the devtools wouldn't let me copy it, sorry.
> 
> Screenshot 2025-08-04 at 16 27 31
> This issue doesn't show up on Android devices.

Can you please tell me the steps to reproduce this issue? Thanks.

LazyAfternoons:
> > > I have encountered an error during remote presentations on an iOS 18.5 device, which seems to be related to a navigator problem, here the details
> > > ```
> > > The action 'GO_BACK' was not handled by any navigator.
> > > 
> > > Is there any screen to go back to?
> > > 
> > > This is a development-only warning and won't be shown in production.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Below there's a stack trace of the error, the devtools wouldn't let me copy it, sorry.
> > > Screenshot 2025-08-04 at 16 27 31
> > > This issue doesn't show up on Android devices.
> > 
> > 
> > Can you please tell me the steps to reproduce this issue? Thanks.
> 
> 1. Start from a clean install on iOS
> 2. Setup the wallet and obtain PID and credentials
> 3. Go to a web portal and start a remote presentation
> 4. After scanning the QR code, instead of returning to the _Presentations_ section, the app warps back to the main menu and the error popup appears (The presentation flow terminates anyway).

Thanks. This issue is not related to this PR but rather to the camera screen logic which doesn't prevent multiple scans, thus `navigation.goBack()` gets called several times. I pushed a fix which prevents this from happening while also removing the persistence of the presentation state.

manuraf (approved):

PR #270, opened 2025-08-04
chore: [SIW-2824] Update eslint config  

LazyAfternoons:
#### List of Changes
- Remove the eslint config from the example app `package.json`; 
- Update the eslint config in the main `package.json`;
- Ignore the `lib` directory as the eslint gets stuck there if it's generated;

#### Motivation and Context
This PR updates the eslint config dependency which was mistakenly added to the example app `package.json` and was a few versions behind in the root `package.json`.

Run `yarn lint` which should work as expected.

#### Checklist:


adelloste (approved):
LGTM

PR #274, opened 2025-08-07
fix: [SIW-2859] Format type for backward-compatible status assertion

gispada:
#### List of Changes

- The status assertion functions now accept the legacy `vc+sd-jwt` as format
- The constant LEGACY_SD_JWT has been created to better distinguish the legacy format

#### Motivation and Context

For backward-compatibility reasons the status assertion functions need to work with legacy credentials. Currently the format type is extracted dynamically from the `obtainCredential`'s return type, that only encompasses the new formats.

The issue is only at the types level, as the compiled code properly handles the old format.

#### Checklist:


ChrisMattew (approved):
LGTM 💪🏻

adelloste (approved):
LGTM

PR #279, opened 2025-09-08
fix: [SIW-2934] Refactor trust federation `buildTrustChain` chain flow

ale-mazz:
#### List of Changes
- Refactored the `buildTrustChain` function to optimize performance when a non-federated entity is detected.

#### Motivation and Context
- This refactor was necessary to reduce processing overhead when a non-federated entity is encountered.

- Tested the flow with all entities in the Trust Federation screen (PROD mode) in the example app. Verified that no regressions occurred.


src/trust/build-chain.ts
```diff
@@ -307,7 +294,8 @@ export async function buildTrustChain(
     const federationList = await getFederationList(federationListEndpoint, {
       appFetch,
     });
-
+    console.log("Federation List:", federationList);
+    console.log("RP url", relyingPartyEntityBaseUrl);
```
gispada, line 298: Did you forget these logs?

gispada (approved):

src/trust/build-chain.ts
ale-mazz: Thanks for the catch, removed!

gispada (approved):
LGTM

PR #347, opened 2026-04-09
fix: [SIW-4130] Improve credential catalogue schema

ale-mazz:
#### List of Changes

This PR aligns the credential catalogue types and mappers
#### Motivation and Context
This change was required to align our schema with the IPZS catalogue deployed to pre-prod

Try to obtain the catalogue using the example app
#### Checklist:


src/credentials-catalogue/v1.3.3/mappers.ts
```diff
@@ -40,8 +40,14 @@ export const mapToCredentialsCatalogue = createMapper<
     }): ApiAuthenticSource => {
       const as = authSourcesById.get(id);
       assert(as, `AS ${id} must be present in the Authentic Source Registry`);
-      const { ipa_code, ...rest } = as.organization_info;
-      return { id, organization_code: ipa_code, ...rest };
+      const { ipa_code, organization_name_l10n_id, ...rest } =
+        as.organization_info;
+      return {
+        id,
+        organization_name: organization_name_l10n_id,
```
gispada: If we drop the `_l10n_id` suffix, how can we know which strings need to be translated? If we assume that localized strings are downloaded separately, we should have a way to match those translations with the corresponding keys.

src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
```diff
@@ -58,16 +73,32 @@ export const DigitalCredential = z.object({
   credential_type: z.string(),
   legal_type: z.string(),
   name: z.string(),
-  description: z.string(),
+  description: z.string().optional(),
+  restriction_policy: z
+    .object({
+      allowed_wallet_ids: z.array(z.string()),
```
gispada: Is it really necessary to expose the allowed Wallet and Issuer IDs in the public API?

ale-mazz: Not really. Removed, thanks!

src/credentials-catalogue/v1.3.3/mappers.ts
ale-mazz: That was my bad. I kept both fields in the public API and correctly mapped them in the mapper. Let me know if this is how you would do it. Thanks!

gispada (approved):
LGTM

PR #350, opened 2026-04-21
feat: [SIW-4180] Add translation fetching logic for credentials catalogue

ale-mazz:
#### List of Changes

 - Added `fetchTranslations` to the `CredentialsCatalogueApi` interface (v1.3.3)
 - Added `LocalizationInfo` and `CatalogueTranslations` types to the shared API layer
 - Extended `DigitalCredentialsCatalogue` with optional `localization` and `as_localization` fields
 - Updated the v1.3.3 mapper to pass localization metadata through to the output
 - Added `fetchLocaleBundle` utility to fetch locale bundle JSON files from the registry
 - Fixed `restriction_policy.allowed_wallet_ids` and `allowed_issuer_ids` to be optional in the v1.3.3 schema
 - Added example app support: thunk, reducer state, and a new button in the Credentials Catalogue screen

 #### Motivation and Context

 The IT-Wallet registry (v1.3.3) exposes localization metadata in both the credential catalogue and
 the authentic sources registry. These point to locale bundle files (flat key→value JSON) that
 resolve the `_l10n_id` keys used throughout the catalogue into human-readable strings.

 This change adds the ability to fetch and merge those bundles for one or more locales, returning
 a single `Record<locale, Record<key, string>>` object ready for use in the UI.

 - Unit tests cover the main behaviors: merged output per locale, AS keys winning on conflict,
   silently skipping locales not in `available_locales`, and graceful handling of missing
   localization metadata
 - Manually tested against the pre environment via the example app

 #### Checklist:

### DIFF
--- a/src/credentials-catalogue/v1.0.0/index.ts
+++ b/src/credentials-catalogue/v1.0.0/index.ts
@@ -1,12 +1,6 @@
 import type { CredentialsCatalogueApi } from "../api";
 import { fetchAndParseCatalogue } from "./fetch-and-parse-catalogue";
-import { IoWalletError } from "../../utils/errors";
 
 export const CredentialsCatalogue: CredentialsCatalogueApi = {
   fetchAndParseCatalogue,
-  fetchTranslations: () => {
-    throw new IoWalletError(
-      "fetchTranslations is not supported in v1.0.0 (no localization metadata)"
-    );
-  },
 };
--- a/src/credentials-catalogue/v1.0.0/index.ts
+++ b/src/credentials-catalogue/v1.0.0/index.ts
@@ -1,6 +1,12 @@
 import type { CredentialsCatalogueApi } from "../api";
 import { fetchAndParseCatalogue } from "./fetch-and-parse-catalogue";
+import { IoWalletError } from "../../utils/errors";
 
 export const CredentialsCatalogue: CredentialsCatalogueApi = {
   fetchAndParseCatalogue,
+  fetchTranslations: () => {
+    throw new IoWalletError(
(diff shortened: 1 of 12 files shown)


src/credentials-catalogue/v1.0.0/index.ts
```diff
@@ -1,6 +1,12 @@
 import type { CredentialsCatalogueApi } from "../api";
 import { fetchAndParseCatalogue } from "./fetch-and-parse-catalogue";
+import { IoWalletError } from "../../utils/errors";
 
 export const CredentialsCatalogue: CredentialsCatalogueApi = {
   fetchAndParseCatalogue,
+  fetchTranslations: () => {
+    throw new IoWalletError(
```
gispada:
```suggestion
    throw new UnimplementedFeatureError("fetchTranslations", "1.0.0")
```

However, do you think we can safely determine whether we can call `fetchTranslations` in v1.0? For instance we only call it if the catalogue includes `localization`?

Otherwise another approach is to make `fetchTranslations` optional in the interface, so the caller should always check if the method exists.

ale-mazz: Thanks for pointing this out. I think making `fetchTranslations` optional on the interface is the cleanest approach. It reflects that `v1.0.0` simply doesn't have this feature, and TypeScript enforces the guard.

ale-mazz: Fix is here https://github.com/pagopa/io-react-native-wallet/pull/350/changes/cdeb3c3ceef4fbc1d24d2e677cdb7504a88fbd4e

gispada (approved):
LGTM

PR #356, opened 2026-05-15
chore: [SIW-3992] Improve status list verification function

gispada:
#### List of Changes

- The status is returned as a string from `verifyAndParseStatusList`
- Modified the example app to test the status list of SD-JWT credentials and the WUA

#### Motivation and Context

This PR modifies the `verifyAndParseStatusList` function to return more meaningful information, and adds a few test scenarios in the example app to get the status list, while keeping the status assertion when supported.


ale-mazz (approved):
LGTM

PR #359, opened 2026-05-22
feat: [SIW-4335] add getStatusL10nIds utility

MonicaRungi-EY:
#### List of Changes
  - Added `getStatusL10nIds(statusBit, allowedStates)` utility function to `src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts`
  - Exported the `AllowedState` type (inferred from the Zod schema) from the catalogue API
  - Re-exported both from `src/credentials-catalogue/api/index.ts` to make them accessible via the `CredentialsCatalogue` namespace
  - Added unit tests in `src/credentials-catalogue/api/__tests__/getStatusL10nIds.test.ts`

#### Motivation and Context
  The credential catalogue exposes an `allowed_states` array for each credential type, where each entry carries the localisation keys needed to display a
  human-readable title and description for that state:

  ```json
  {
    "0x00": "VALID",
    "title_l10n_id": "mDL.VALID.title",
    "description_l10n_id": "mDL.VALID.description"
  }
  ```

  At runtime, `verifyAndParseStatusList` returns a `statusBit` (e.g. `"0x00"`, `"0x0B"`) representing the credential's current state read from the status list
  JWT. There was no utility to bridge the two: given a `statusBit`, find the matching `allowed_states` entry and retrieve its l10n keys.

  This function is the missing link that allows consumers to resolve a runtime status bit into localised UI strings, without having to re-implement the lookup
  (including case-insensitive matching) themselves.

  Note: the comparison is case-insensitive because `verifyAndParseStatusList` produces uppercase hex (`"0x0B"`) while catalogue keys use lowercase (`"0x0b"`).

Unit tests cover the following cases:
  - Match with a lowercase `statusBit` (`"0x00"`) — nominal path
  - Match with an uppercase `statusBit` (`"0x0B"`) — verifies case-insensitive lookup
  - `statusBit` not present in `allowedStates` → returns `undefined`
  - Empty `allowedStates` array → returns `undefined`
  - Mixed array containing plain strings (v1.0.0 catalogue format) alongside `AllowedState` objects → strings are skipped without errors
  - Array containing only plain strings → returns `undefined`

 N/A — no UI changes.

#### Checklist:

### DIFF
--- a/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
+++ b/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
@@ -1,11 +1,6 @@
 import * as z from "zod";
 import { UnixTime } from "../../utils/zod";
 
-export type GetStatusL10nIds = (
-  statusBit: string,
-  credentialConfig: DigitalCredential
-) => { titleL10nId: string; descriptionL10nId: string } | undefined;
-
 export const LocalizationInfo = z.object({
   available_locales: z.array(z.string()),
   base_uri: z.string(),
--- a/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
+++ b/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
@@ -27,7 +27,6 @@ const AllowedState = z
   })
   .catchall(z.string());
 
-
 /**
  * Given a statusBit (e.g. "0x00", "0x0B") and a DigitalCredential from the
  * catalogue, returns the matching l10n IDs or undefined if not found.
--- a/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
+++ b/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
@@ -1,6 +1,11 @@
 import * as z from "zod";
 import { UnixTime } from "../../utils/zod";
 
+export type GetStatusL10nIds = (
+  statusBit: string,
+  credentialConfig: DigitalCredential
+) => { titleL10nId: string; descriptionL10nId: string } | undefined;
+
 export const LocalizationInfo = z.object({
   available_locales: z.array(z.string()),
   base_uri: z.string(),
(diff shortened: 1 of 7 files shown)


src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts
```diff
@@ -27,6 +27,31 @@ const AllowedState = z
...
 
+export type AllowedState = z.infer<typeof AllowedState>;
+
+/**
+ * Given a statusBit (e.g. "0x00", "0x0B") and the allowed_states array
+ * from the catalogue, returns the matching l10n IDs or undefined if not found.
+ * The comparison is case-insensitive to handle uppercase statusBit values
+ * returned by verifyAndParseStatusList against lowercase keys in the catalogue.
+ */
+export const getStatusL10nIds = (
```
gispada: I would change the signature to `getStatusL10nIds(statusBit: string, credentialConfig: DigitalCredential)` so we don't need to extract `allowedStates` in the consumer app.

You can extract the type from [here](https://github.com/pagopa/io-react-native-wallet/blob/master/src/credentials-catalogue/api/DigitalCredentialsCatalogue.ts#L87).

gispada (changes requested):
The API files should only contain type/interfaces, not implementations.

`getStatusL10nIds` should stay in a separate file under a specific version folder, or if it does not break any version you can move it to `credentials-catalogue/common/`.

We should only add the `getStatusL10nIds` type to the API, then export its implementation in the relevant `v1.*/index.ts`.

src/credentials-catalogue/api/index.ts
```diff
@@ -48,11 +50,23 @@ export interface CredentialsCatalogueApi {
...
+
+  /**
+   * Given a statusBit (e.g. "0x00", "0x0B") and a DigitalCredential from the
+   * catalogue, returns the matching l10n IDs or undefined if not found.
+   * The comparison is case-insensitive to handle uppercase statusBit values
+   * returned by verifyAndParseStatusList against lowercase keys in the catalogue.
+   *
+   * @since 1.0.0
+   */
+  getStatusL10nIds: GetStatusL10nIds;
```
gispada: Let's be consistent with the rest of the file and declare `getStatusL10nIds` inline, without importing the type from DigitalCredentialsCatalogue.

src/credentials-catalogue/common/get-status-l10n-ids.ts
```diff
@@ -0,0 +1,27 @@
...
+  type DigitalCredential,
+} from "../api/DigitalCredentialsCatalogue";
+
+/**
+ * Given a statusBit (e.g. "0x00", "0x0B") and a DigitalCredential from the
+ * catalogue, returns the matching l10n IDs or undefined if not found.
+ * The comparison is case-insensitive to handle uppercase statusBit values
+ * returned by verifyAndParseStatusList against lowercase keys in the catalogue.
+ */
+export const getStatusL10nIds = (
```
gispada: You can avoid redeclaring the function signature:

```suggestion
export const getStatusL10nIds: CredentialsCatalogueApi["getStatusL10nIds"] = (
```

gispada (approved):
LGTM

PR #361, opened 2026-05-29
feat(example): [SIW-4380] Add batch issuance test scenario

gispada:
#### List of Changes

- Added a test scenario for batch issuance of the new proof of age credential.

#### Motivation and Context

This PR allows testing the batch issuance in the example app.

Tested with the v1.3 EAA Issuer in the pre env. Please note that proof of age fails mdoc validation for a missing claim. The batch issuance, however, should work as expected.


ale-mazz (approved):
LGTM

PR #365, opened 2026-06-09
chore: [SIW-0000] Accept external function to fetch the redirect uri when completing issuance user auth

gispada:
#### Checklist:


ale-mazz (approved):
LGTM

PR #367, opened 2026-06-15
chore: [SIW-0000] Make `x5c` required in Wallet Instance Attestation 1.3 header

gispada:
#### List of Changes

This PR modifies the 1.3.3 Wallet Instance Attestation zod type to require `x5c`. This change results in a failure when parsing a 1.0.0 attestation, allowing to properly distinguish versions by their JWT structure.

It also makes `integration_endpoint` and `user_information_l10n_id` optional in the Authentic Source data capability object.


ale-mazz (approved):
LGTM

PR #368, opened 2026-06-16
chore: [SIW-4431] Integrate DCQL query external evaluation into UA flow

DavideOnano:
#### List of Changes

- Exposed the evaluated `dcql_query` result to SDK consumers
- Updated EAA authorization APIs to receive `evaluatedDcqlQuery` instead of the PID tuple
- Applied the change to issuance `v1.0.0` and `v1.3.3`
- Updated the example app and issuance README

#### Motivation and Context

IO needs the `dcql_query` result before completing user authorization, in order to read `requiredDisclosures` and show only the requested PID claims in the Trust Issuer screen.

Previously the SDK evaluated the query internally during `completeUserAuthorization`, so those disclosures were not available to the caller early enough.

Verified with the example app that `requiredDisclosures` are available before `completeUserAuthorization` by adding logs before `await wallet.CredentialIssuance.completeUserAuthorizationWithFormPostJwtMode`:
1. Obtain a WI & WIA
2. Obtain PID
3. Go to Credential and choose an EAA
4. Verify that the dqcl query result exists and are as expected
#### Checklist:


example/src/utils/credential.ts
```diff
@@ -12,6 +12,7 @@ import last from "lodash/last";
 import appFetch from "../utils/fetch";
 import { DPOP_KEYTAG, regenerateCryptoKey } from "./crypto";
 import type { CryptoContext } from "@pagopa/io-react-native-jwt";
+import type { DcqlQuery } from "dcql";
```
gispada: Let's not import from `dcql` directly in the consumer app.
```ts
type DcqlQuery = Parameters<
  RemotePresentation.RemotePresentationApi["evaluateDcqlQuery"]
>[0];
```
Or we could re-export the type.

src/credential/issuance/api/03-complete-user-authorization.ts
```diff
@@ -4,11 +4,17 @@ import type {
   AuthorizationResult,
 } from "../../../utils/auth";
 import type { RequestObject } from "../../../credential/presentation/api";
+import type { EvaluateDcqlQueryApi } from "../../../credential/presentation/api/06-evaluate-dcql-query";
+import type { Out } from "../../../utils/misc";
 import type { IssuerConfig } from "./IssuerConfig";
 
+export type EvaluatedDcqlQueryResult = Out<
```
gispada:
```suggestion
export type EvaluatedDcqlQuery = Out<
```

src/credential/issuance/README.md
```diff
@@ -96,7 +96,7 @@ When the credential is different than an eID, the flow requires the user to pres
 
 ## Authentication through credentials (Form Post JWT Mode) - v1.0
```
gispada, line 97: I would update the documentation for **Authentication through credentials (Query  Mode) - v1.3** as well.

gispada (approved):
LGTM

PR #369, opened 2026-06-23
chore: [SIW-4510] Split status list functions to allow getting TSL with from a uri

mastro993:
#### List of Changes

- Refactored `StatusListApi` interface: reorganized the process of fetching, verifying, and reading status list tokens.
- Created three core methods: split the functionality into `get`, `verifyAndParse`, and a new `getStatus` method specifically for extracting status bits.
- Introduced `getByUri`: added a new method to allow direct URI fetching.
- Updated type definitions: improved types and return values to ensure clarity and correctness.

#### Motivation and Context

In the IO app, Status List tokens are persisted in their JWT payload. We need to find a way to keep them up to date by reusing the io-react-native-wallet library functions. Since the library only has methods to fetch a Status List from a credential JWT, we are splitting the function to allow fetching a Status List from a URI.

- Obtain a credential
- Obtain status list for credential
- Verify that everything is working

#### Checklist:


src/credential/status/v1.3.3/03-get-status.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!status) {
+    throw new IoWalletError(
+      `Unsupported credential status bit: ${formatStatusBit(statusBit)}`
+    );
+  }
+
+  return {
+    status,
+    statusBit: formatStatusBit(statusBit),
```
gispada: Since we are here, I was thinking we can rename `statusBit`: the status might be defined by more than one bit when "bits" > 1. Maybe `statusValue` or `rawStatus` would be more precise?

mastro993: Totally agree. I've renamed `statusBit` into `rawStatus` [here](https://github.com/pagopa/io-react-native-wallet/pull/369/commits/23b6c06635547c40ef3bd00a644b3707fdf0c76b)

gispada (approved):
LGTM

PR #373, opened 2026-07-03
chore: [SIW-4573] Omit claims without display property from parsed credential

gispada:
#### List of Changes

- Filtered `credentialConfig.claims` in mdoc and sd-jwt parsers.

#### Motivation and Context

This PR implements https://github.com/italia/eid-wallet-it-docs/pull/1167 and properly handles the now optional `display` property in each claim definition. 

#### Checklist:


ale-mazz (approved):
LGTM

PR #382, opened 2026-07-28
chore: Migrate to nx monorepo with expo

LazyAfternoons:
#### List of Changes
- Replace the monorepo setup with `nx`;
- Replace the example app with a expo app and update dependencies which must be updated to work with the new `react-native` version;
- Resolve the circular dependency issues in the example app which caused runtime crashes on expo;
- Minor updates to the code due to the new `react` version;
- Add `io-react-native-wallet` as library in the monorepo;
- Temporary remove `debug-logging-server` which will be reintroduced later to avoid too many changes in this PR.

#### Motivation and Context
This PR migrates this monorepo to `nx` to align it with `devex` standards. Most of the changes are files that are either being move or being deleted due to expo CGN which doesn't require the native projects to be commited. Static checks will fail until this gets merged and we change the branch protection rules. Statics can be run locally by running `pnpm validate` or manually through a filter. This PR doesn't change the way we release `io-react-native-wallet` package or we generate changelog, as this requires broader changes which are out of scope. 

This is the first step to a centralized monorepo for the IT-Wallet implementation and related libraries which will be gradually moved in this repository.

Test the example apps on both platforms, everything should work as before.
The required setup and commands can be found in the updated `README.md` files. After the merge we must run a release of the `io-react-native-wallet` package to check if everything is still working fine.

#### Checklist:


gispada (approved):
LGTM, tested on Android.
Great work

PR #388, opened 2026-09-10
chore: Anonymize fiscal codes in `IssuerResponseError` reason

mastro993:
#### List of Changes

- `IssuerResponseError` runs `reason` through `anonymizeString` before `super()`, so both `reason` and the serialized `Error.message` replace Italian fiscal codes (codice fiscale, including omocodia) with 16 asterisks
- added `anonymizeString` in `src/utils/string.ts`: walks strings, arrays, and nested objects; other values pass through. The fiscal-code regex is built per call so a shared `/g` `lastIndex` cannot leak
- `ResponseErrorBuilder.buildFrom` passes `reason` and `statusCode` into the specialized error constructor instead of spreading the original `Error` instance
- `UnexpectedStatusCodeError`, `WalletProviderResponseError`, and `RelyingPartyResponseError` are unchanged

#### Motivation and Context

Credential issuer error payloads can include a codice fiscale in `reason`. App code logs and displays that field. This PR strips fiscal codes from `IssuerResponseError` only, which is the type consumers see for issuer HTTP failures.

Unit tests:
- `anonymizeString`: one or many fiscal codes in a string, nested objects, arrays, primitives, strings with no fiscal code
- `IssuerResponseError`: string `reason`, object `reason`, `reason` with no fiscal code

#### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,11 +143,11 @@
     ]
   },
   "dependencies": {
-    "@pagopa/io-wallet-oauth2": "1.5.4",
-    "@pagopa/io-wallet-oid-federation": "1.5.4",
-    "@pagopa/io-wallet-oid4vci": "1.5.4",
-    "@pagopa/io-wallet-oid4vp": "1.5.4",
-    "@pagopa/io-wallet-utils": "1.5.4",
+    "@pagopa/io-wallet-oauth2": "1.5.6",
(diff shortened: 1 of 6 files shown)


gispada (approved):
LGTM

ale-mazz (approved):

